PR #43, opened 2023-01-07
Implement a function to return schema vocabularies

jviotti:


include/jsontoolkit/json/rapidjson/iterators.h
```diff
@@ -13,7 +13,7 @@ inline auto value(const rapidjson::Value::Member &member) -> const Value & {
   return member.value;
 }
 
-inline auto key(rapidjson::Value::Member &member) -> std::string {
+inline auto key(const rapidjson::Value::Member &member) -> std::string {
```
jviotti, line 16: Silly mistake on me

PR #55, opened 2023-01-28
Check vocabularies at every step in `subschema_iterator()`

jviotti:
To account for bundled schemas with different vocabularies in use within
the same top-level schema.


jviotti:
TODO: Test traversing a multi-metaschema schema

PR #57, opened 2023-02-03
Implement a flat subschema iterator

jviotti:


include/jsontoolkit/jsonschema/walker.h
```diff
@@ -25,30 +25,37 @@ enum class schema_walker_strategy_t {
...
                    const schema_resolver_t &resolver,
-                   const std::string &metaschema) -> void;
+                   const std::string &metaschema, const std::size_t level)
+      -> void;
   auto walk_schema(const Value &schema, const schema_walker_t &walker,
                    const schema_resolver_t &resolver,
-                   const std::string &metaschema) -> void;
+                   const std::string &metaschema, const std::size_t level)
+      -> void;
   std::vector<std::reference_wrapper<const Value>> subschemas;
```
jviotti, line 57: Note to self: we want `flat_subschema_iterator` to be read-write. The `SchemaWalker` class hardcodes a vector of const JSON values, so we will probably need to turn `SchemaWalker` into a templated class that takes the vector element type as a template argument. `subschema_iterator` would pass `const Value` and `flat_subschema_iterator` would pass `Value`.

jviotti: Done!

PR #91, opened 2023-04-15
Implement a sample contrib cURL resolver

jviotti:


jviotti:
TODO to self:

- [ ] Link to this as an example of a class-based resolver in the docs
- [ ] Use this resolver to power the JSON Schema related contrib CLIs

PR #290, opened 2023-10-07
Add hyper-schema walker support to 2019-09 and 2020-12

jviotti:


jviotti:
This is incomplete. See keywords from `links.json` too

PR #300, opened 2023-10-10
Implement a basic URI class based on `uriparser`

jviotti:


jviotti:
TODO: Add docs

PR #405, opened 2023-12-21
Enable every test from the JSON Schema Referencing Suite

jviotti:


jviotti:
See https://github.com/python-jsonschema/referencing-suite/pull/38

PR #541, opened 2024-04-15
Add base and relative pointer to compiler context

jviotti:

### DIFF
--- a/src/jsonschema/compile.cc
+++ b/src/jsonschema/compile.cc
@@ -33,6 +33,7 @@ auto compile_subschema(
     const auto &keyword{entry.pointer.back().to_property()};
     for (auto &&step : context.compiler(
              {keyword, context.schema, entry.vocabularies, entry.value,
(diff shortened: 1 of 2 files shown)


src/jsonschema/compile.cc
```diff
@@ -89,12 +92,25 @@ auto compile(const JSON &schema, const SchemaWalker &walker,
...
+
+  const auto &new_schema{get(context.value, pointer)};
   return compile_subschema(
-      {context.keyword, get(context.value, pointer), context.vocabularies,
-       context.value, context.evaluation_path.concat(pointer),
-       context.instance_location, context.frame, context.references,
-       context.walker, context.resolver, context.compiler,
-       context.default_dialect});
+      {context.keyword, new_schema,
+       vocabularies(new_schema, context.resolver, entry.dialect).get(),
```
jviotti, line 109: TODO: We need a `vocabularies` overload that just takes a dialect

PR #564, opened 2024-04-19
Corrected some typos

Era-cell:
In the assert() - The function in this PR asserts to false, I guess its intended to be true always as per other examples.


jviotti (approved):
Thanks a lot! Great catch

jviotti:
I wonder why GitHub Actions didn't trigger... I think I know what the problem might be though...

jviotti:
I think I fixed it. Next time it should work

PR #600, opened 2024-04-27
Implement Draft 4 `patternProperties`

jviotti:


jviotti:
TODO: Start testing with traces. Primarily annotation values now

PR #642, opened 2024-05-07
Add a "contains" compiler assertion step

jviotti:


jviotti:
Actually this is wrong. We need an equality check instead

PR #688, opened 2024-05-10
Fix bundling directly-recursive schemas

jviotti:

### DIFF
--- a/src/jsonschema/bundle.cc
+++ b/src/jsonschema/bundle.cc
@@ -54,10 +54,16 @@ auto bundle_schema(sourcemeta::jsontoolkit::JSON &root,
       .wait();
 
   for (const auto &[key, reference] : references) {
+    // TODO: Canonicalize destinations directly during framing instead
+    const auto canonical_destination{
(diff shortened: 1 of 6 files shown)


test/jsonschema/jsonschema_bundle_2020_12_test.cc
```diff
@@ -370,3 +378,32 @@ TEST(JSONSchema_bundle_2020_12, taken_definitions_entry) {
 
   EXPECT_EQ(document, expected);
 }
+
+TEST(JSONSchema_bundle_2020_12, recursive) {
```
jviotti, line 382: TODO: Get this test on every supported dialect

src/jsonschema/bundle.cc
```diff
@@ -54,10 +54,15 @@ auto bundle_schema(sourcemeta::jsontoolkit::JSON &root,
       .wait();
 
   for (const auto &[key, reference] : references) {
+    const auto canonical_destination{
```
jviotti, line 57: TODO: Investigate why the official test suite still chokes, even with this

PR #804, opened 2024-06-12
fix: add .cache to .gitignore

tony-go:


jviotti (approved):
Looks good but what creates `.cache` out of curiosity? Never seen it before

jviotti:
Haha weird, but I guess it doesn't work adding it

PR #806, opened 2024-06-12
feat(uri): add `user_info()` API

### LINKED ISSUE #805: Add a `[[nodiscard]] auto URI::userinfo() const -> std::optional<std::string_view>` method

tony-go:
Fixes #805

### DIFF
--- a/src/uri/include/sourcemeta/jsontoolkit/uri.h
+++ b/src/uri/include/sourcemeta/jsontoolkit/uri.h
@@ -299,12 +299,12 @@ public:
   ///
   /// const sourcemeta::jsontoolkit::URI uri{"https://user:@host"};
   /// assert(uri.user_info().has_value());
-  /// assert(uri.user_info().value() == "user");
+  /// assert(uri.user_info().value() == "user:);
   /// ```
   ///
   /// As mentioned in RFC 3986, the format "user:password" is deprecated.
   /// Applications should not render as clear text any data after the first
-  /// colon. Cf. https://tools.ietf.org/html/rfc3986#section-3.2.1
+  /// colon. See https://tools.ietf.org/html/rfc3986#section-3.2.1
   [[nodiscard]] auto user_info() const -> std::optional<std::string_view>;
 
 private:
--- a/src/uri/include/sourcemeta/jsontoolkit/uri.h
+++ b/src/uri/include/sourcemeta/jsontoolkit/uri.h
@@ -298,14 +298,14 @@ public:
   /// #include <cassert>
   ///
   /// const sourcemeta::jsontoolkit::URI uri{"https://user:@host"};
-  /// assert(uri.user_info().has_value());
-  /// assert(uri.user_info().value() == "user:);
+  /// assert(uri.userinfo().has_value());
+  /// assert(uri.userinfo().value() == "user:);
   /// ```
   ///
   /// As mentioned in RFC 3986, the format "user:password" is deprecated.
   /// Applications should not render as clear text any data after the first
   /// colon. See https://tools.ietf.org/html/rfc3986#section-3.2.1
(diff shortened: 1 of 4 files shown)


test/uri/uri_user_info_test.cc
```diff
@@ -0,0 +1,22 @@
...
+  EXPECT_EQ(uri.user_info().value(), "user:");
+}
+
+// Note: Using the format "user:password" is depreacted.
+// But we would like to support it for users that would still use it.
+TEST(URI_user_info, depreacted_example) {
+  sourcemeta::jsontoolkit::URI uri{"http://user:password@host:80/"};
+  EXPECT_TRUE(uri.user_info().has_value());
+  EXPECT_EQ(uri.user_info().value(), "user:password");
+}
```
tony-go, line 17: I wonder if we should not be strict on this? or at least print a warning.

jviotti: I think it's fine. Even if it's deprecated, is part of the RFC. The warning you put in the docs should be enough!

test/uri/uri_user_info_test.cc
```diff
@@ -0,0 +1,22 @@
+
+#include <gtest/gtest.h>
+#include <sourcemeta/jsontoolkit/uri.h>
+
+TEST(URI_user_info, working_example) {
```
jviotti, line 5: Can you test user info with a URN? (if it even makes sense there?)

jviotti: Also maybe test it with a relative URI as well, just for extra coverage, which might come handy when we do our own parser

tony-go: > Can you test user info with a URN? (if it even makes sense there?)

Regarding my first reading on URNs and tags, it seems userinfo does not have to deal with.

tony-go: > lso maybe test it with a relative URI as well

Relative URI represents only path right? I don't think we could use user_info, scheme or port in this case.

jviotti: On relative URIs, exactly! You can to assert that the userinfo is empty

jviotti: > Regarding my first reading on URNs and tags, it seems userinfo does not have to deal with.

yes, I confirm! Maybe still add a test that checks that for a URN, the userinfo returns a non-value?

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -288,8 +288,25 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
...
+  ///
+  /// const sourcemeta::jsontoolkit::URI uri{"https://user:@host"};
+  /// assert(uri.user_info().has_value());
+  /// assert(uri.user_info().value() == "user:);
+  /// ```
+  ///
+  /// As mentioned in RFC 3986, the format "user:password" is deprecated.
+  /// Applications should not render as clear text any data after the first
+  /// colon. See https://tools.ietf.org/html/rfc3986#section-3.2.1
+  [[nodiscard]] auto user_info() const -> std::optional<std::string_view>;
```
jviotti, line 308: Can we call this `userinfo`? The RFC refers to it as such, even though it uses hyphens for other parts of the grammar. Maybe worth sticking with that convention?

jviotti: Hm, I don't have strong opinions actually. This looks good too. I don't know

jviotti (approved):
I left some minor remaining comments but looks good. Feel free to merge when ready!

src/uri/include/sourcemeta/jsontoolkit/uri.h
tony-go: done

PR #810, opened 2024-06-13
refactor(uri): internalize uri structure

tony-go:
This PR aims to avoid redudant parsing and then limiting dependency with `uriparser`.


src/uri/uri.cc
```diff
@@ -89,6 +92,38 @@ URI::URI(URI &&other)
...
+
+  uri_parse(this->data, &this->internal->uri);
+
+  const UriPathSegmentA *segment{this->internal->uri.pathHead};
+  if (!segment) {
+    this->parsed = true;
+    return;
+  }
+
+  // URNs and tags have one segement bu default
```
tony-go, line 111: typoe

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -151,9 +153,10 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
   /// const sourcemeta::jsontoolkit::URI
   /// uri{"https://www.sourcemeta.com/foo/bar"};
   /// assert(uri.path().has_value());
-  /// assert(uri.path().value() == "/foo/bar");
+  /// assert(uri.path().value().at(0) == "foo");
+  /// assert(uri.path().value().at(1) == "bar");
   /// ```
-  [[nodiscard]] auto path() const -> std::optional<std::string>;
+  [[nodiscard]] auto path() const -> std::optional<std::span<std::string>>;
```
jviotti, line 159: Maybe I'm totally wrong, but I *think* you can do const here:

```suggestion
  [[nodiscard]] auto path() const -> std::optional<std::span<const std::string>>;
```

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -291,6 +294,9 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
   static auto from_fragment(std::string_view fragment) -> URI;
 
 private:
+  bool parsed = false;
```
jviotti, line 297: Do we really need to have this? As far as I can see (to which I agree) all parsing should happen on the constructors, so we can guarantee the URI is parsed when the constructor is done?

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -301,6 +307,9 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
   // points to fragments of it.
   // We keep this as const as this class is immutable
   std::string data;
+
+  mutable std::vector<std::string> components;
```
jviotti, line 311: Maybe call this `path` or `path_components`? Otherwise its not very clear at first sight that its about paths?

src/uri/uri.cc
```diff
@@ -193,7 +211,11 @@ auto URI::recompose_without_fragment() const -> std::optional<std::string> {
   // Path
   const auto result_path{this->path()};
   if (result_path.has_value()) {
-    result << result_path.value();
+    auto seprator = (this->is_urn() || this->is_tag()) ? "" : "/";
```
jviotti, line 214: Typo?

src/uri/uri.cc
```diff
@@ -252,7 +274,11 @@ auto URI::canonicalize() -> URI & {
...
```
jviotti, line 277: Same here

src/uri/include/sourcemeta/jsontoolkit/uri.h
tony-go: It was my assumption before I remember that we have this issue: #811 

If we potentially re-set stuff, we need to parse it again right?

Also I noticed that on the end of this method we reparse stuff again: 
- https://github.com/sourcemeta/jsontoolkit/blob/main/src/uri/uri.cc#L295-L299
- https://github.com/sourcemeta/jsontoolkit/blob/main/src/uri/uri.cc#L270-L272

src/uri/include/sourcemeta/jsontoolkit/uri.h
tony-go: Yeah :)

src/uri/include/sourcemeta/jsontoolkit/uri.h
tony-go: > Assuming you do all parsing in the constructor, then you don't need it here?

Yeah. According what we decide here: https://github.com/sourcemeta/jsontoolkit/pull/810#discussion_r1638331668

src/uri/include/sourcemeta/jsontoolkit/uri.h
jviotti: The idea is that we only parse one. Its a bit tricky given the intermediate state with `uriparser`, but what I'm thinking about is:

- We only perform parsing at the constructors and we populate a set of members with the result. For example, we have a member that is the host, a member that is the fragment, a member that is the path, etc.
- At this point ideally we don't store the PIMPL private `uriparser` struct anymore as a member
- The setters then just interact with the pre-parsed members, never parsing again

jviotti: The ultimate goal here is to never keep a `uriparser` struct around anymore. We only do it on demand when we need to (and ideally we stop needing it at some point). For example, if we parse into a set of nice members on the class, then we might need to re-construct a `uriparser` instance on the fly for relative resolution or recomposing, and then throw it away, etc.

src/uri/include/sourcemeta/jsontoolkit/uri.h
tony-go: gcc is not happy with that haha!

src/uri/include/sourcemeta/jsontoolkit/uri.h
tony-go: > We only perform parsing at the constructors and we populate a set of members with the result. For example, we have a member that is the host, a member that is the fragment, a member that is the path, etc.

So we could parse once, then every method should not reparse but use existing members.

Which means that this should not happen?

- https://github.com/sourcemeta/jsontoolkit/blob/main/src/uri/uri.cc#L295-L299
- https://github.com/sourcemeta/jsontoolkit/blob/main/src/uri/uri.cc#L270-L272

tony-go: I'm trying to understand, how much could I apply this final vision to the refactor of URI::path which is consumed by URI::canonicalize (that reparse on the end).

jviotti: @tony-go I think its fine for the code you highlighted to happen, just that instead of affecting a `uriparser` class member directly, it would:

1. Construct back, on the fly, a `uriparser` struct based on the new class members
2. Do whatever it needs with `uriparser` to this temporary instance
3. Decode back into the new members
4. Discard the `uriparser` struct

jviotti: The core goal here is that we stop storing a `uriparser` struct as a class member, and instead put our own C++-friendly class members. Something like this:

```c++
private:
  std::optional<std::string> scheme;
  std::optional<std::string> userinfo;
  std::optional<std::string> host;
  // etc
```

That way, instead of making `uriparser` an intrinsic part of the class, we just use it on demand when needed and then discard it. That gets us closer to eventually getting rid of `uriparser`

tony-go: Yeah we are aligned. My whole point is either we add these properties and I can do uri_parse once only in the constructor.

jviotti: Yeah, let's do it!

src/uri/uri.cc
```diff
@@ -89,6 +92,54 @@ URI::URI(URI &&other)
...
+      this->path_components_.push_back(std::string{part.value()});
+    } else {
+      while (segment) {
+        const auto part{uri_text_range(&segment->text)};
+        assert(part.has_value());
+        this->path_components_.push_back(std::string{part.value()});
+        segment = segment->next;
+      }
+    }
+  }
```
tony-go, line 126: It seems that for some reasons, `pathHead` points to the host.

Like in this test case:

```cpp
TEST(URI_path, without_scheme) {
  const sourcemeta::jsontoolkit::URI uri{"example.com/foo"};
  EXPECT_SPAN_EQ(uri.path().value(), {"foo"});
}
```

Which fails:

```text
/Users/tonygo/oss/jsontoolkit/test/uri/uri_path_test.cc:11: Failure
Expected equality of these values:
  span[i]
    Which is: "example.com"
  *it
    Which is: "foo"
Span element 0 does not match expected value
```

Note: I just catch the bug, the test is not pushed on this PR.

cc @jviotti

tony-go: I'll try to reproduce it on a standalone bug, to see what I missed here. It could be a silly side effect or something like that...

tony-go: I was able to reproduce it: 

```cpp
#include <iostream>
#include <string>
#include <uriparser/Uri.h>

auto uri_text_range(const UriTextRangeA *const range)
    -> std::optional<std::string_view> {
  if (range->afterLast == nullptr) {
    return std::nullopt;
  }

  return std::string_view{range->first,
                          static_cast<std::string_view::size_type>(
                              range->afterLast - range->first)};
}

auto print_path(const std::string uri) {
  UriUriA uri_struct;
  const char *error_position;
  uriParseSingleUriA(&uri_struct, uri.c_str(), &error_position);

  auto seg = uri_struct.pathHead;
  while (seg != nullptr) {
    const auto text = uri_text_range(&seg->text);
    if (text.has_value()) {
      std::cout << text.value() << std::endl;
    }
    seg = seg->next;
  }

  uriFreeUriMembersA(&uri_struct);
}

int main() {
  const std::string input_without_scheme = "example.com/foo/";
  const std::string input_with_scheme = "https://example.com/foo/bar/baz";

  print_path(input_without_scheme);
  print_path(input_with_scheme);
}
```

```
example.com
foo

foo
bar
baz
```

In one case we have the host, not in the other. Maybe it is expected. 🤔

tony-go: My 2 cents is that if the parser does not have enough context (the scheme) he probably deduce that is a path.

tony-go: I guess that I could update `normalize`/`canonicalize` to handle this case properly. For instance if I don't have a scheme in the URL then a `/` should not prefix the first path element.

jviotti: Hmmm to be honest I'm not sure. Before making assumptions or tweaking `normalize`/`canonicalize`, can you research what other popular parsers (maybe in JS, etc) do? Maybe it is indeed the expectation and we would have learnt something new.

jviotti: Or maybe even opening an issue on `uriparser` asking as well. URIs are very obscure, so I wouldn't be surprised if this feels weird for us but its actually compliant behavior, etc

tony-go: Yeah, I'll :)

tony-go: Quite interesting: https://datatracker.ietf.org/doc/html/rfc3986#section-4.5

jviotti: Yeah, very interesting. I also checked on the JSON Schema side and I think the current behavior is intended. i.e. if you just get `example.com/foo`, then it should be considered a path of `{ example.com, foo }` and we should not try to be smart and detect that `example.com` is the host

tony-go: Yeah for sure. What I was thinking about was to handle "/" for this specific case (we only have exception for URN and tag so far)

Lets go!

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -318,6 +324,15 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
   // points to fragments of it.
   // We keep this as const as this class is immutable
   std::string data;
+
+  mutable std::vector<std::string> path_components_;
```
tony-go, line 328: Reminder: try to remove it.

src/uri/uri.cc
```diff
@@ -70,25 +73,83 @@ struct URI::Internal {
...
 URI::URI(URI &&other)
     : data{std::move(other.data)}, internal{std::move(other.internal)} {
+  this->parsed = other.parsed;
+  this->path_components_ = std::move(other.path_components_);
+  this->scheme_ = std::move(other.scheme_);
+  this->userinfo_ = std::move(other.userinfo_);
+  this->host_ = std::move(other.host_);
+  this->port_ = std::move(other.port_);
+  this->fragment_ = std::move(other.fragment_);
+  this->query_ = std::move(other.query_);
```
tony-go, line 100: I wonder if we should not have a URIElement struct to abstract of all these.

src/uri/uri.cc
```diff
@@ -245,31 +350,12 @@ auto URI::canonicalize() -> URI & {
...
-
-  // Fragment
-  const auto result_fragment{this->fragment()};
-  if (result_fragment.has_value() && !result_fragment.value().empty()) {
-    result << '#' << result_fragment.value();
-  }
-
-  this->data = result.str();
-  uriFreeUriMembersA(&this->internal->uri);
-  uri_parse(this->data, &this->internal->uri);
```
tony-go, line 272: No more parsing :) I should remove it in other places too.

test/uri/uri_canonicalize_test.cc
```diff
@@ -68,10 +68,16 @@ TEST(URI_canonicalize, example_10) {
   EXPECT_EQ(uri.recompose(), "urn:example:schema");
 }
 
-TEST(URI_canonicalize, example_11) {
+TEST(URI_canonicalize, example_relative_1) {
   sourcemeta::jsontoolkit::URI uri{"../foo"};
   uri.canonicalize();
-  EXPECT_EQ(uri.recompose(), "/foo");
+  EXPECT_EQ(uri.recompose(), "foo");
```
tony-go, line 74: ⚠️ @jviotti

test/uri/uri_recompose_test.cc
```diff
@@ -18,7 +18,7 @@ TEST(URI_recompose, urn) {
 
 TEST(URI_recompose, https_with_empty_fragment) {
   const sourcemeta::jsontoolkit::URI uri{"https://example.com/foo#"};
-  EXPECT_EQ(uri.recompose(), "https://example.com/foo#");
+  EXPECT_EQ(uri.recompose(), "https://example.com/foo");
```
tony-go, line 21: ⚠️ @jviotti

test/uri/uri_recompose_test.cc
```diff
@@ -33,5 +33,5 @@ TEST(URI_recompose, no_scheme) {
 
 TEST(URI_recompose, empty_fragment) {
   const sourcemeta::jsontoolkit::URI uri{"#"};
-  EXPECT_EQ(uri.recompose(), "#");
+  EXPECT_EQ(uri.recompose(), "");
```
tony-go, line 36: ⚠️ @jviotti

src/uri/uri.cc
```diff
@@ -70,25 +73,83 @@ struct URI::Internal {
...
+    this->path_components_.clear();
+    this->scheme_ = std::nullopt;
+    this->userinfo_ = std::nullopt;
+    this->host_ = std::nullopt;
+    this->port_ = std::nullopt;
+    this->fragment_ = std::nullopt;
+    this->query_ = std::nullopt;
+    this->parsed = false;
+    uriFreeUriMembersA(&this->internal->uri);
+  }
```
tony-go, line 117:
```suggestion
  assert(!this->parsed)
```

Once I remove all re-parsing from the whole class.

I don't know if I should do it now or in another PR.

src/uri/uri.cc
```diff
@@ -111,57 +172,82 @@ auto URI::is_fragment_only() const -> bool {
...
+  if (result_fragment.has_value() && !result_fragment.value().empty()) {
+    result << '#' << result_fragment.value();
+  }
 
-auto URI::recompose() const -> std::string {
-  return uri_to_string(&this->internal->uri);
+  return result.str();
 }
 
 auto URI::recompose_without_fragment() const -> std::optional<std::string> {
```
tony-go, line 253: Should be factorized properly with `recompose`

src/uri/uri.cc
jviotti: > I don't know if I should do it now or in another PR.

Up to you!

src/uri/uri.cc
jviotti: I think it's fine as it is. Very simple and direct

test/uri/uri_canonicalize_test.cc
jviotti: I think this one is OK.

test/uri/uri_recompose_test.cc
jviotti: I think we should address these differences. In theory `https://example.com/foo#` and `https://example.com/foo` are different URIs. They may point to the same thing, but that should happen only when doing `canonicalize()`. Otherwise changing the URI that the user is giving us might be a recipe for disaster.

test/uri/uri_recompose_test.cc
jviotti: I think we should fix this one too. `"#"` is URI with an empty fragment, while `""` is an empty URI, which is not the same

test/uri/uri_canonicalize_test.cc
jviotti: This one should be fixed too. I think this indicates that storing path components in a vector without the slash might not do, as we might lose some information? Maybe the path components here should look like this (i.e. keeping the slash as part of the component?):

```c++
[ "/foo" ]
```

tony-go: I don't know so far. I'll think about it.

But if add the slash in `recompose` then we will end up having `/example.com/foo` for the input `example.com/foo`.

tony-go: Now the thing is that we use `uriparser` for the parsing, and we won't have the information if the slash was there or not. We could check `this->data` to see if the there is slash or not before the path.

test/uri/uri_recompose_test.cc
tony-go: Done

test/uri/uri_canonicalize_test.cc
jviotti: Yeah, sounds tricky. The thing is that if we get a URI and recompose, it should ideally get you back the exact same thing you put in there, and any updates to clean it up, etc should only happen under canonicalisation.

Maybe keeping the path component as a vector doesn't help much?

tony-go: > Maybe keeping the path component as a vector doesn't help much?

yeah we should keep a string a server a span on demand!

test/uri/uri_path_test.cc
```diff
@@ -9,25 +9,25 @@ TEST(URI_path, https_example_url_no_path) {
...
   EXPECT_TRUE(uri.path().has_value());
-  EXPECT_EQ(uri.path().value(), "/foo/bar/baz");
+  EXPECT_EQ(uri.path().value(), "foo/bar/baz");
 }
 
 TEST(URI_path, relative_multi) {
   const sourcemeta::jsontoolkit::URI uri{"../foo/bar"};
   EXPECT_TRUE(uri.path().has_value());
-  EXPECT_EQ(uri.path().value(), "/../foo/bar");
+  EXPECT_EQ(uri.path().value(), "../foo/bar");
```
tony-go, line 30: Hmmm let me check that

test/uri/uri_resolve_from_test.cc
```diff
@@ -28,16 +28,16 @@ TEST(URI_resolve_from, change_fragment) {
...
+//   relative.resolve_from_if_absolute(base);
+//   EXPECT_EQ(relative.recompose(), "https://foobar.com/baz");
+// }
+//
+// TEST(URI_resolve_from_if_absolute, relative_base) {
+//   const sourcemeta::jsontoolkit::URI base{"../foo"};
+//   sourcemeta::jsontoolkit::URI relative{"../baz"};
+//   relative.resolve_from_if_absolute(base);
+//   EXPECT_EQ(relative.recompose(), "../baz");
+// }
```
tony-go, line 43: and revert that.

test/uri/uri_path_test.cc
tony-go: What I don't really understand with this is that we have this:

```cpp
TEST(URI_path, without_scheme) {
  const sourcemeta::jsontoolkit::URI uri{"example.com/foo"};
  EXPECT_EQ(uri.path().value(), "example.com/foo");
}
``` 

and we would to have this on the other side:

```
TEST(URI_path, relative_multi) {
  const sourcemeta::jsontoolkit::URI uri{"../foo/bar"};
  EXPECT_TRUE(uri.path().has_value());
  EXPECT_EQ(uri.path().value(), "/../foo/bar");
}
```

I identifier both of the cases as:
- NON URN
- NON tag
- WHICH have not scheme.

So I think we should have a take here.

CC @jviotti

test/uri/uri_canonicalize_test.cc
tony-go, line 74: Regarding what I see [here](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.4): 

```
If the input buffer begins with a prefix of "../" or "./",
then remove that prefix from the input buffer; otherwise,

I think we should keep it like that.
```

jviotti: @tony-go You are reading a section under reference resolution. But on `canonicalize` I think that's fine too :+1:

test/uri/CMakeLists.txt
```diff
@@ -23,4 +23,4 @@ target_link_libraries(sourcemeta_jsontoolkit_uri_unit
   PRIVATE sourcemeta::jsontoolkit::uri)
 set_target_properties(sourcemeta_jsontoolkit_uri_unit
   PROPERTIES FOLDER "JSON Toolkit/URI")
-add_test(NAME JSON COMMAND sourcemeta_jsontoolkit_uri_unit --gtest_brief=1)
+add_test(NAME URI COMMAND sourcemeta_jsontoolkit_uri_unit --gtest_brief=1)
```
jviotti, line 26: Good catch!

test/uri/uri_path_test.cc
jviotti: Yeah, good point. I think how you have them now (with the current changes to these tests) looks good to me.

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -318,6 +323,15 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
   // points to fragments of it.
   // We keep this as const as this class is immutable
   std::string data;
+
+  std::optional<std::string> path_;
```
jviotti, line 327: A bit of a nit pick, but would it make sense to re-order these members to reflect the actual ordering in a URI? For example, start with the scheme, then userinfo, etc? I think it'd make it easier to read

test/uri/uri_path_test.cc
```diff
@@ -9,25 +9,25 @@ TEST(URI_path, https_example_url_no_path) {
 TEST(URI_path, https_example_url_slash) {
   const sourcemeta::jsontoolkit::URI uri{"https://example.com/"};
   EXPECT_TRUE(uri.path().has_value());
-  EXPECT_EQ(uri.path().value(), "/");
+  EXPECT_EQ(uri.path().value(), "");
```
jviotti, line 12: @tony-go I think the test at the bottom of this file (for `../foo/bar`) is good after the update, but the rest should have the leading `/`, as that's technically part of the path.

As a consequence, right now you cannot distinguish between i.e. `https://example.com/` and `https://example.com`, as the paths would both be the empty string.

Can you make sure that if the paths are not relative, they indeed start with `/`?

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -308,6 +310,9 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
   [[nodiscard]] auto userinfo() const -> std::optional<std::string_view>;
 
 private:
+  bool parsed = false;
```
jviotti, line 313: Can we add a TODO item to remove this `parsed` member eventually? I think there are ways to do it without it and that will look a bit cleaner

jviotti (changes requested):

test/uri/uri_path_test.cc
jviotti: But this is overall why I wish there was a really good existing URI test suite we could pull...

jviotti: At least https://github.com/hyperjump-io/uri, which is a compliant URI implementation I trust (from another TSC member of JSON Schema) seems to keep the leading `/`

src/uri/include/sourcemeta/jsontoolkit/uri.h
tony-go: Yeah I should do the same with the parse method.

src/uri/include/sourcemeta/jsontoolkit/uri.h
tony-go: Yeah I don't like it, but until we have migrate all other method to use our internal structure I prefer to keep it. Lets add the todo.

test/uri/uri_path_test.cc
tony-go: we should see if there is a leading `/` during parsing and storing it.

tony-go: For the test suite, lets do it into a subsequent PR.

jviotti: What do you mean for the test suite?

tony-go: Including the test suit in jsontoolkit project

tony-go: > we should see if there is a leading `/` during parsing and storing it.

i still not resolved it btw. I probably need a few more sessions 😅

jviotti: > Including the test suit in jsontoolkit project

What I meant is that there is no existing URI test suite out there we can use that is good, and that's the challenge in all of this. We have to compare and research a lot more. It'd be great if there was a very solid and very compliant URI parsing suite out there

tony-go: Sorry, I got confused when you said

> I wish there was a really good existing URI

I thought it was locally in the project 🤪

src/uri/uri.cc
```diff
@@ -63,32 +65,141 @@ static auto uri_parse(const std::string &data, UriUriA *uri) -> void {
   uri_normalize(uri);
 }
 
+static auto
+canonicalize_path(const std::string &path) -> std::optional<std::string> {
```
tony-go, line 69: I would really like to unit test this properly!

jviotti (approved):
Awesome job!

PR #880, opened 2024-07-29
deps: add google benchmark

tony-go:
This PR aims to add google benchmark as deps:
- Write a dummy benchmark
- Run it on CI


tony-go:
Let see if I did not break anything :)

Then I could add it to the CI :)

.github/workflows/ci.yml
```diff
@@ -123,6 +125,9 @@ jobs:
           # See https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html
           UBSAN_OPTIONS: print_stacktrace=1
 
+      - name: Benchmarks
+        run: cmake --build ./build --config Release --verbose --target run_compiler_benchmark
```
jviotti, line 129:
```suggestion
        run: cmake --build ./build --config Release --verbose --target benchmark
```

We can keep it simple? I think we will expand on more benchmarks than just the compiler at some point. For example, we might end up profiling the JSON class as a consequence

.github/workflows/ci.yml
```diff
@@ -123,6 +125,9 @@ jobs:
           # See https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html
           UBSAN_OPTIONS: print_stacktrace=1
 
+      - name: Benchmarks
```
jviotti, line 128: Minor thing, but in the rest of the GitHub Action definition we don't put names in the steps, as they are short enough to understand?

CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ option(JSONTOOLKIT_JSONSCHEMA "Build the JSON Toolkit JSON Schema library" ON)
 option(JSONTOOLKIT_JSONPOINTER "Build the JSON Toolkit JSON Pointer library" ON)
 option(JSONTOOLKIT_JSONL "Build the JSON Toolkit JSONL library" ON)
 option(JSONTOOLKIT_TESTS "Build the JSON Toolkit tests" OFF)
+option(JSONTOOLKIT_BENCHMARKS "Build the JSON Toolkit benchmarks" OFF)
```
jviotti, line 15:
```suggestion
option(JSONTOOLKIT_BENCHMARK "Build the JSON Toolkit benchmark" OFF)
```

CMakeLists.txt
```diff
@@ -106,4 +107,10 @@ if(JSONTOOLKIT_TESTS)
       add_subdirectory(test/packaging)
     endif()
   endif()
+
+  if(JSONTOOLKIT_BENCHMARKS)
+    find_package(benchmark REQUIRED)
```
jviotti: I would have expected this to be:

```suggestion
    find_package(GoogleBenchmark REQUIRED)
```

DEPENDENCIES
```diff
@@ -14,3 +14,4 @@ jsonschema-draft0 https://github.com/json-schema-org/json-schema-spec 7ea575aef8
 referencing-suite https://github.com/python-jsonschema/referencing-suite 61c4cc202b1e96ed5adcaf4842a595f68d659212
 jsonschema-test-suite https://github.com/json-schema-org/JSON-Schema-Test-Suite c2badb1298a8698f86dadf1aea7b44b3a894e5ac
 uriparser https://github.com/uriparser/uriparser bfe94f6e54d0abb5afa7bb0411940b7242cb835a
+benchmark https://github.com/google/benchmark 378fe693a1ef51500db21b11ff05a8018c5f0e55
```
jviotti, line 17: Can we call this `googlebenchmark` to match how we do `googletest` above?

benchmark/compiler/CMakeLists.txt
jviotti, line 1: I think you might need to introduce a better directory structure for us to expand into more benchmarks. I think we will have more than one benchmark for the various modules shipped in this library. For example:

```
benchmark/<module>/<name>
```

Like `benchmark/jsonschema/compiler-properties`

benchmark/compiler/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+add_executable(compiler_benchmark basic.cc)
```
jviotti, line 1: Also, we tend to fully qualify target names, as this library if for downstream consumption and we might get some name collisions at some point? For example `sourcemeta_jsontoolkit_jsonschema_benchmark_xxxxxx`?

jviotti: We could eventually have nice macros at Noa to hide all of these things

benchmark/compiler/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+
+target_link_libraries(compiler_benchmark benchmark::benchmark)
+target_link_libraries(compiler_benchmark benchmark::benchmark_main)
+target_link_libraries(compiler_benchmark sourcemeta::jsontoolkit::json)
+target_link_libraries(compiler_benchmark sourcemeta::jsontoolkit::jsonschema)
+
+add_custom_target(run_compiler_benchmark
```
jviotti, line 8: I would expect one target that runs ALL of the benchmarks in the project?

benchmark/compiler/basic.cc
jviotti, line 1: Because `compiler` is not a module in this project, maybe have `benchmark/jsonschema/compiler_basic.cc`?

vendor/benchmark.mask
jviotti, line 1: Can you mask all of the unnecessary files from here: I guess only `CMakeLists.txt` from there is necessary. 

Screenshot 2024-07-29 at 9 52 02 AM

.github/workflows/ci.yml
tony-go: Yeah is simpler Indeed :)

I do back and forth between both approach:
- simplicity on ones side
- specificity on the other as you rarely want to run all locally, and even on the CI, you'll prolly look for a specific segment..

.github/workflows/ci.yml
tony-go: I definetly prefer names. For instance `cmake --build ./build ...` looks quite long and unreadable to me hahah

benchmark/compiler/CMakeLists.txt
tony-go: Yeah you are right :)

benchmark/compiler/CMakeLists.txt
tony-go: Yeah we want to have a target to run all of the benchmarks... but I did not create this one initially for all :)

benchmark/compiler/basic.cc
tony-go: yeah as you said above :)

vendor/benchmark.mask
tony-go: These ones were hidden in my IDE, nice catch :) I'll add then to the .mask file

CMakeLists.txt
tony-go: Now I wonder, why do we have plural for test?

benchmark/compiler/CMakeLists.txt
tony-go: I'll create one for all :)

.github/workflows/ci.yml
```diff
@@ -123,6 +125,9 @@ jobs:
...
+        run: cmake --build ./build --config Release --verbose --target run_benchmark
```
jviotti, line 129:
```suggestion
        run: cmake --build ./build --config Release --verbose --target benchmark
```

We are executing a custom target, so we are by definition "running" it?

.github/workflows/ci.yml
jviotti: Right. Maybe. Can you avoid the title now for consistency? You can add the titles for all the steps in another PR if we want and we discuss there?

benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_subdirectory(jsonschema)
+
+add_custom_target(run_benchmark
+    COMMAND sourcemeta_jsontoolkit_benchmark_jsonschema
+    DEPENDS sourcemeta_jsontoolkit_benchmark_jsonschema
+    WORKING_DIRECTORY ${CMAKE_PROJECT_DIR}
+    COMMENT "Running benchmark...")
+
```
jviotti, line 8: Empty line here

benchmark/jsonschema/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_executable(sourcemeta_jsontoolkit_benchmark_jsonschema compiler_basic.cc)
+
+target_link_libraries(sourcemeta_jsontoolkit_benchmark_jsonschema benchmark::benchmark)
```
jviotti, line 3: You are missing the visibility attribute here. I guess all of them should be `PRIVATE`

benchmark/jsonschema/compiler_basic.cc
```diff
@@ -0,0 +1,25 @@
...
+    const sourcemeta::jsontoolkit::JSON schema{
+        sourcemeta::jsontoolkit::parse(R"JSON({
+      "$schema": "https://json-schema.org/draft/2019-09/schema",
+      "properties": {
+        "foo": { "type": "string" },
+        "bar": { "type": "integer" }
+      }
+    })JSON")};
+
+    const auto compiled_schema{sourcemeta::jsontoolkit::compile(
```
jviotti, line 17: Google Benchmark has some macros that can prevent the compiler optimising these away. Can you take look? Otherwise, because you are not using this variable, some compiler configurations will get rid of the entire thing and we will not be benchmarking anything

benchmark/jsonschema/compiler_basic.cc
```diff
@@ -0,0 +1,25 @@
+#include <benchmark/benchmark.h>
+
+#include <sourcemeta/jsontoolkit/json.h>
+#include <sourcemeta/jsontoolkit/jsonschema.h>
+
+static void Compiler_Basic(benchmark::State &state) {
+  for (auto _ : state) {
+    const sourcemeta::jsontoolkit::JSON schema{
```
jviotti, line 8: I think this JSON parsing can be outside the measured part, as we are only benchmarking the compiler

vendor/googlebenchmark/.ycm_extra_conf.py
jviotti, line 1: Can you mask this one?

vendor/googlebenchmark/.pre-commit-config.yaml
jviotti, line 1: Can you mask this one?

benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_subdirectory(jsonschema)
+
+add_custom_target(run_benchmark
+    COMMAND sourcemeta_jsontoolkit_benchmark_jsonschema
+    DEPENDS sourcemeta_jsontoolkit_benchmark_jsonschema
+    WORKING_DIRECTORY ${CMAKE_PROJECT_DIR}
```
jviotti, line 6: I don't think you need to specify a working directory here, as all the benchmarks we will use are working directory agnostic? We can let CMake use the current binary directory as a default

benchmark/jsonschema/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
...
+target_link_libraries(sourcemeta_jsontoolkit_benchmark_jsonschema benchmark::benchmark_main)
+target_link_libraries(sourcemeta_jsontoolkit_benchmark_jsonschema sourcemeta::jsontoolkit::json)
+target_link_libraries(sourcemeta_jsontoolkit_benchmark_jsonschema sourcemeta::jsontoolkit::jsonschema)
+
```
jviotti, line 7: Empty line here

benchmark/jsonschema/compiler_basic.cc
```diff
@@ -0,0 +1,25 @@
...
+      }
+    })JSON")};
+
+    const auto compiled_schema{sourcemeta::jsontoolkit::compile(
+        schema, sourcemeta::jsontoolkit::default_schema_walker,
+        sourcemeta::jsontoolkit::official_resolver,
+        sourcemeta::jsontoolkit::default_schema_compiler)};
+  }
+}
+BENCHMARK(Compiler_Basic);
```
jviotti, line 23: For readability, I think you could add one blank line above this one, and maybe remove the blank line below?

.github/workflows/ci.yml
tony-go: We are conflicting with `benchmark` itself haha :)

tony-go:
```cmake
target_link_libraries(sourcemeta_jsontoolkit_benchmark_jsonschema benchmark::benchmark)
```

benchmark/jsonschema/compiler_basic.cc
tony-go: yeah sorry for all of this, TBH I just wrote it to have something that "runs" 😅

.github/workflows/ci.yml
jviotti: Ah, crap. What about `benchmark_all`? I think that would be more CMake idiomatic

benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+add_subdirectory(jsonschema)
```
jviotti, line 1: Now that I realise, you are activating this file if `JSONTOOLKIT_BENCHMARK` is set, but then for each of these subdirectories, you need to check the relevant component variables. For example, here, you should only enable this directory if `JSONTOOLKIT_JSONSCHEMA`, etc? 

That means we also need to gracefully handle what happens if you enable benchmarks but no actual components. I guess you can just exit with a `FATAL_ERROR` but you could also create a custom target that prints a useful string like "no benchmarks detected" and that does an exit 1?

jviotti:
Very nice! Just one last comment. This was very useful to see how it's done. Now that we know how it goes, it'd be nice to upstream some concise macros to Noa, so we don't have to deal with all of this boilerplate

tony-go:
> Very nice! Just one last comment. This was very useful to see how it's done. Now that we know how it goes, it'd be nice to upstream some concise macros to Noa, so we don't have to deal with all of this boilerplate

Let's explore it a bit more and take experience with it.

benchmark/CMakeLists.txt
tony-go: > you should only enable this directory if JSONTOOLKIT_JSONSCHEMA, etc?

that should be enough no? I mean regarding this:

> That means we also need to gracefully handle what happens if you enable benchmarks but no actual components. I guess you can just exit with a FATAL_ERROR but you could also create a custom target that prints a useful string like "no benchmarks detected" and that does an exit 1?

tony-go: I let a todo. I think we could explore that later and focus on tackling the performance issue first.

jviotti (approved):
Looks good!

PR #885, opened 2024-07-30
Introduce a method to get a relative URI out of absolute ones

jviotti:

### DIFF
--- a/src/uri/uri.cc
+++ b/src/uri/uri.cc
@@ -408,6 +408,48 @@ auto URI::resolve_from(const URI &base) -> URI & {
   }
 }
 
+auto URI::relative_to(const URI &base) -> URI & {
+  if (!this->is_absolute() || !base.is_absolute()) {
+    return *this;
+  }
+
+  UriUriA result;
+  URI copy{*this};
+
+  // According to the docs, we only need to free on success
+  // See
+  // https://uriparser.github.io/doc/api/latest/Uri_8h.html#a20cc7888b62700d6cb7e7896647b0d5d
+  switch (uriRemoveBaseUriMmA(&result, &copy.internal->uri, &base.internal->uri,
+                              URI_FALSE, nullptr)) {
+    case URI_SUCCESS:
+      break;
+    default:
+      throw URIError{"Could not resolve URI relative to the given base"};
+  }
+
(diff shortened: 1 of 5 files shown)


src/uri/uri.cc
```diff
@@ -408,6 +408,33 @@ auto URI::resolve_from(const URI &base) -> URI & {
   }
 }
 
+auto URI::relative_to(const URI &base) -> URI & {
+  UriUriA result;
+
+  // According to the docs, we only need to free on success
+  // See
+  // https://uriparser.github.io/doc/api/latest/Uri_8h.html#af948b48ec0749192f1a944b044d5726f
+  switch (uriRemoveBaseUriMmA(&result, &this->internal->uri,
```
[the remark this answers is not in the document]
jviotti: For sure, which is what you were working towards on the last PR. I think we will be able to get rid of it at some point soon.

test/uri/uri_relative_to_test.cc
```diff
@@ -0,0 +1,72 @@
...
+  EXPECT_EQ(uri.recompose(), "foo");
+}
+
+TEST(URI_relative_to, relative_absolute_1) {
+  const sourcemeta::jsontoolkit::URI base{"foo/bar"};
+  sourcemeta::jsontoolkit::URI uri{"https://www.example.com/foo"};
+  uri.relative_to(base);
+  EXPECT_EQ(uri.recompose(), "https://www.example.com/foo");
+}
+
```
tony-go, line 66: I think We should have better name than this :)

tony-go (approved):
One nitpick!

test/uri/uri_relative_to_test.cc
jviotti: What do you mean?

tony-go: Like explicitly mentioning what we are testing ‘relative_schemeless_uri’ or whatever make sense :)

jviotti: Ah, yeah, fair enough. We should also do a lot more testing once we implement this purely ourselves. I wonder what edge cases `uriparser` is already sorting out here that we are not aware of

PR #887, opened 2024-07-31
feat(uri): add setter for path

tony-go:
Part of #811


test/uri/uri_path_test.cc
```diff
@@ -1,64 +1,163 @@
...
             "bowtie.report,2023-11:referencing-suite-tag-uris-id");
 }
 
-TEST(URI_path, without_scheme) {
+TEST(URI_path_getter, without_scheme) {
   const sourcemeta::jsontoolkit::URI uri{"example.com/foo"};
   EXPECT_EQ(uri.path().value(), "example.com/foo");
 }
+
+// Setter
```
tony-go, line 68: Most of focus should be here!

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -157,6 +157,19 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
   /// ```
   [[nodiscard]] auto path() const -> std::optional<std::string>;
 
+  /// Set the path part of the URI. Foe example:
```
jviotti, line 160:
```suggestion
  /// Set the path part of the URI. For example:
```

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -157,6 +157,19 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
...
+  ///
+  /// ```cpp
+  /// #include <sourcemeta/jsontoolkit/uri.h>
+  /// @include <cassert>
```
jviotti, line 164:
```suggestion
  /// #include <cassert>
```

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -157,6 +157,19 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
...
+  ///
+  /// sourcemeta::jsontoolkit::URI uri{"https://www.sourcemeta.com"};
+  /// uri.path("/foo/bar");
+  /// assert(uri.path().has_value());
+  /// assert(uri.path().value() == "/foo/bar");
+  /// ```
+  auto path(std::string_view path) -> URI &;
```
jviotti, line 171: Given we store the data as an `std::string`, I think you might be better off taking `const std::string &`

jviotti: On this, I think you should also have an overload for `std::string &&` for performance reasons

src/uri/uri.cc
```diff
@@ -249,6 +249,29 @@ auto URI::path() const -> std::optional<std::string> {
...
+
+  const auto query_pos = path.find('?');
+  if (query_pos != std::string::npos) {
+    // Keep only whast we have before ?
+    path = path.substr(0, query_pos);
+  }
+
+  const auto fragment_pos = path.find('#');
+  if (fragment_pos != std::string::npos) {
+    // Keep only whast we have before #
```
jviotti, line 266:
```suggestion
    // Keep only what we have before #
```

src/uri/uri.cc
```diff
@@ -249,6 +249,29 @@ auto URI::path() const -> std::optional<std::string> {
   return path_;
 }
 
+auto URI::path(std::string_view path) -> URI & {
+  if (path.empty()) {
+    this->path_ = std::nullopt;
+    return *this;
+  }
+
+  const auto query_pos = path.find('?');
```
jviotti, line 258: This logic makes me think you might want to parse the `path` as a proper `URI`, and then move its `path` member to the class?

test/uri/uri_path_test.cc
```diff
@@ -1,64 +1,163 @@
...
+  uri.path("/foo%20bar#fragment");
+  EXPECT_EQ(uri.path().value(), "/foo%20bar");
+}
+
+TEST(URI_path_setter, set_path_with_query_and_fragment) {
+  sourcemeta::jsontoolkit::URI uri{
+      "https://example.com/old?query=value#fragment"};
+  uri.path("/new");
+  EXPECT_EQ(uri.path().value(), "/new");
+  EXPECT_EQ(uri.recompose(), "https://example.com/new?query=value#fragment");
```
jviotti, line 123: I like the `recompose` check here. Maybe you should check `recompose()` in all setter tests?

src/uri/include/sourcemeta/jsontoolkit/uri.h
tony-go: > Given we store the data as an `std::string,` I think you might be better off taking `const std::string &`

I have two concerns with that:
- I am not sure to understand the link that there is between how we store data and the interface itself.
- But using `const` I cannot modify `path` and I have to create copy, then I was wondering about potential perf implication. Does that make sense?

tony-go: > On this, I think you should also have an overload for std::string && for performance reasons

I struggled with that haha!

Let me do it in another PR, when I can focus 100% on this.

src/uri/uri.cc
tony-go: Is that approach not "too much"? In term of performance :)

tony-go: I mean I really like it, principaly because it will simplify the code ;)

src/uri/include/sourcemeta/jsontoolkit/uri.h
jviotti: I think on your points above, that is the reason C++ setters usually have both copy and move overloads. Then the developer can pick the one that works best on their specific circumstance!

src/uri/uri.cc
jviotti: The thing is that you are parsing anyway to some extent, and that "some extent" might not cover all edge cases you want to cover, so if we are already doing some sort of parsing, maybe we do all of it?

In any case, let's make it works first, elegantly, covering all edge cases, etc and then we can profile & optimise later on if needed

jviotti:
Not sure if here or in another PR, but as you implement setters or any new functionality in URI, it might be worth searching uses of URI in other modules, like `jsonschema` and see if they can be refactored to use the new functionality. I remember leaving lots of TODOs for improvements on URIs that we planned (though not sure if for path setters though)

tony-go:
> Not sure if here or in another PR, but as you implement setters or any new functionality in URI, it might be worth searching uses of URI in other modules, like `jsonschema` and see if they can be refactored to use the new functionality. I remember leaving lots of TODOs for improvements on URIs that we planned (though not sure if for path setters though)

Yeah let me try :)

src/uri/uri.cc
```diff
@@ -249,6 +249,22 @@ auto URI::path() const -> std::optional<std::string> {
...
+
+  const auto parsed_path = URI{std::string{path}};
+  const bool has_leading_slash = path.front() == '/';
+  const std::string path_to_set = parsed_path.path().value();
+
+  this->path_ = has_leading_slash ? path_to_set.substr(1) : path_to_set;
+  return *this;
+}
+
+auto URI::path(std::string &&path) -> URI & { return this->path(path); }
```
jviotti, line 266: I don't think you are actually implementing move semantics. You are calling back to the copy overload, thus forcing a copy. You will need the implementation here to be more elaborate and likely NOT call the other one. 

I think you will end up repeating some of the boilerplate, but passing the string ownership to the `URI` constructor, and then moving `path` from your temporary URI to the path member in the current class

test/uri/uri_path_test.cc
```diff
@@ -1,64 +1,249 @@
...
+
+// Setter
+// NOTE: we test both std::string&& and std::string_view signatures
+// - for `std::string_view` we use `std::string_view{}`
+// - for `std::string&&` we use `std::string{}`
+
+TEST(URI_path_setter, no_path) {
+  sourcemeta::jsontoolkit::URI uri{"https://example.com"};
+
+  uri.path(std::string_view{"/foo"});
```
jviotti, line 77: This is implicitly trying to move, as you are passing an rvalue. To really test a copy, you will probably need to declare a `const` variable with the path string, and then pass that

jviotti (changes requested):

src/uri/uri.cc
tony-go: > You will need the implementation here to be more elaborate and likely NOT call the other one.

Funny because I started with it haha :) 

> I think you will end up repeating some of the boilerplate, but passing the string ownership to the URI constructor, and then moving path from your temporary URI to the path member in the current class

Yeah it makes whole sense.

Move semantics confused me here as I was assuming that it was the string passed as a parameter that was "moved" (`std::move(xxxx)`) and that on our side we don't have to do anything. My reasoning was wrong! 

Note: I realised now that the move constructor I implemented above could have help me to understand it!

test/uri/uri_path_test.cc
tony-go: Yeah I test a case like that at the end of the test suite. But I should generalize it :) Thanks

src/uri/uri.cc
jviotti: I don't remember if I recommended it before, but http://cppmove.com is quite fun to read on this topic

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -157,6 +157,32 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
...
   [[nodiscard]] auto path() const -> std::optional<std::string>;
 
+  /// Set the path part of the URI. For example:
+  ///
+  /// ```cpp
+  /// #include <sourcemeta/jsontoolkit/uri.h>
+  /// #include <cassert>
+  ///
+  /// sourcemeta::jsontoolkit::URI uri{"https://www.sourcemeta.com"};
+  /// uri.path("/foo/bar");
```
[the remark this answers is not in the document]
jviotti: This is still exemplifying a move `rvalue` operation. To truly exemplify copying, you should extract the value into a `const std::string`

src/uri/uri.cc
```diff
@@ -249,6 +249,34 @@ auto URI::path() const -> std::optional<std::string> {
...
+}
+
+auto URI::path(std::string &&path) -> URI & {
+  if (path.empty()) {
+    this->path_ = std::nullopt;
+    return *this;
+  }
+
+  const auto parsed_path = URI{std::move(path)};
+  const std::string path_to_set = parsed_path.path().value();
```
jviotti, line 273: Note that this is still performing a copy. Maybe you can get rid of it?

src/uri/uri.cc
```diff
@@ -249,6 +249,34 @@ auto URI::path() const -> std::optional<std::string> {
...
+  const bool has_leading_slash = path_to_set.front() == '/';
+
+  this->path_ = has_leading_slash ? path_to_set.substr(1) : path_to_set;
```
jviotti, line 276: This is also still performing a copy. At the end of the day, I think this overload is much closer to the copy overload that it should be. Maybe there are ways to make it a proper move?

jviotti: For example: you are already parsing the path as a URI into `parsed_path`. At that point, it is already parsed by your parsing rules, so in theory you could just do this?

```
this->path_ = std::move(parsed_path.path());
```

?

if its `std::nullopt`, I think it would also handle the `if` above too

src/uri/uri.cc
```diff
@@ -249,6 +249,34 @@ auto URI::path() const -> std::optional<std::string> {
...
+  return *this;
+}
+
+auto URI::path(std::string &&path) -> URI & {
+  if (path.empty()) {
+    this->path_ = std::nullopt;
+    return *this;
+  }
+
+  const auto parsed_path = URI{std::move(path)};
```
jviotti, line 272: You probably don't want this to be `const`, as otherwise you still do another copy

src/uri/uri.cc
```diff
@@ -249,6 +249,34 @@ auto URI::path() const -> std::optional<std::string> {
...
 }
 
+auto URI::path(const std::string &path) -> URI & {
+  if (path.empty()) {
+    this->path_ = std::nullopt;
+    return *this;
+  }
+
+  const auto parsed_path = URI{std::string{path}};
+  const std::string path_to_set = parsed_path.path().value();
```
jviotti, line 259: This feels like an unnecessary extra copy.

jviotti: Actually same comment as below. I think it can all be simplified assuming you are already parsing the path in the line above? You can then just copy the `path` member, right?

test/uri/uri_path_test.cc
```diff
@@ -1,64 +1,261 @@
...
   EXPECT_EQ(uri.path().value(), "example.com/foo");
 }
+
+// Setter
+
+TEST(URI_path_setter, no_path) {
+  sourcemeta::jsontoolkit::URI uri{"https://example.com"};
+
+  uri.path(std::string{"/foo"});
+  EXPECT_EQ(uri.path().value(), "/foo");
```
jviotti, line 75: What happens if I set a path that is relative, like `../foo/bar`? You probably want to reject it by throwing an exception, as that case should make use of the URI resolution methods instead

src/uri/uri.cc
tony-go: Oh yeah I should continue to move the std::string as much as I can! Otherwise I'll have a move semantics that contains copy that is totally non sense!

tony-go: By the way this feedback works for both `path` methods

test/uri/uri_path_test.cc
tony-go: hmm let me check!

src/uri/include/sourcemeta/jsontoolkit/uri.h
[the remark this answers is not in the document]
jviotti: This is still getting resolved, but the issue remains!

src/uri/uri.cc
```diff
@@ -249,6 +249,54 @@ auto URI::path() const -> std::optional<std::string> {
   return path_;
 }
 
+auto URI::path(const std::string &path) -> URI & {
+  if (path.empty()) {
+    this->path_ = std::nullopt;
+    return *this;
+  }
+
+  auto is_relative = path.starts_with("./") || path.starts_with("../");
```
jviotti, line 258:
```suggestion
  auto is_relative = path.starts_with(".");
```

Isn't this enough?

jviotti: Also, this variable should be `const`

src/uri/uri.cc
```diff
@@ -249,6 +249,54 @@ auto URI::path() const -> std::optional<std::string> {
...
+  if (is_relative) {
+    throw URIError{
+        "You cannot set a relative path, use URI::resolve_from() instead"};
```
jviotti, line 261:
```suggestion
        "You cannot set a relative path"};
```

I guess we can just explain the issue without specific instructions, otherwise the error message won't make much sense outside of a C++ context

src/uri/uri.cc
```diff
@@ -249,6 +249,54 @@ auto URI::path() const -> std::optional<std::string> {
...
+  }
+
+  auto parsed_path = URI{std::string{path}};
+  if (!parsed_path.path().has_value()) {
+    this->path_ = std::nullopt;
+    return *this;
+  }
+
+  const bool has_leading_slash = parsed_path.path().value().front() == '/';
+  this->path_ = has_leading_slash ? parsed_path.path().value().substr(1)
```
jviotti, line 271: Why do we need this logic? Isn't this enough?

```suggestion
  this->path_ = parsed_path.path();
```

src/uri/include/sourcemeta/jsontoolkit/uri.h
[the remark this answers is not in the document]
tony-go: For some reasons, I skip it, sorry!

src/uri/uri.cc
tony-go: what if I have `.github/workflows/` ?

Fun story! I started with the same approach haha

src/uri/uri.cc
tony-go: Why?

tony-go: let check, but there is an isse with tests where need it! at some point out of the getter, but we decided to not keep it in the `path_` itself!

jviotti: I think you should take a proper look at that. The getter and setter should mirror each other, so a developer would expect that anything you "get" can be "set" without additional modifications, otherwise the setter is doing more work than it should?

tony-go: Yeah I agree! I'll take a deeper look tomorrow :)

test/uri/uri_is_absolute_test.cc
```diff
@@ -15,3 +15,8 @@ TEST(URI_is_absolute, urn) {
   const sourcemeta::jsontoolkit::URI uri{"urn:example:schema"};
   EXPECT_TRUE(uri.is_absolute());
 }
+
+// TEST(URI_is_absolute, slash) {
+//   const sourcemeta::jsontoolkit::URI uri{"/foo"};
```
tony-go, line 20: https://www.rfc-editor.org/rfc/rfc3986#section-4.2 would let me think that this is absolute: 

> A relative reference that begins with a single slash character is termed an absolute-path reference.

Any thoughts @jviotti ?

src/uri/uri.cc
```diff
@@ -240,13 +245,49 @@ auto URI::path() const -> std::optional<std::string> {
...
+
+auto URI::path(const std::string &path) -> URI & {
+  if (path.empty()) {
+    this->path_ = std::nullopt;
+    return *this;
   }
 
-  return path_;
+  const auto is_relative = path.starts_with(".");
+  if (is_relative) {
```
tony-go, line 258:
```suggestion
  const auto is_relative_path = path.starts_with(".");
  if (is_relative_path) {
```

test/uri/uri_is_absolute_test.cc
jviotti: From what I understand (though I'm not a URI expert :P):

- It is a relative URI
- But it is an absolute _path_

So to answer your question, I would expect `is_absolute()` to be false

src/uri/uri.cc
```diff
@@ -240,13 +245,49 @@ auto URI::path() const -> std::optional<std::string> {
...
+    return *this;
+  }
+
+  const auto is_relative_path = path.starts_with(".");
+  if (is_relative_path) {
+    throw URIError{"You cannot set a relative path"};
+  }
+
+  auto parsed_path = URI{std::move(path)};
+  if (!parsed_path.path().has_value()) {
```
jviotti, line 284: I think this whole conditional goes away if you do `this->path_ = parsed_path.path()`?

tony-go: yeah as it is an optional too :)

src/uri/uri.cc
```diff
@@ -240,13 +244,39 @@ auto URI::path() const -> std::optional<std::string> {
...
+    return *this;
   }
 
-  return path_;
+  const auto is_relative_path = path.starts_with(".");
+  if (is_relative_path) {
+    throw URIError{"You cannot set a relative path"};
+  }
+
+  auto parsed_path = URI{std::string{path}};
```
jviotti, line 261: Note that this URI is temporal, so while you copy the input string, you can actually move the URI.

jviotti, line 261: Also, isn't this enough?

```suggestion
  auto parsed_path = URI{path};
```

src/uri/uri.cc
```diff
@@ -240,13 +244,39 @@ auto URI::path() const -> std::optional<std::string> {
...
+  this->path_ = parsed_path.path().value();
```
jviotti, line 262:
```suggestion
  this->path_ = URI{path}.path_;
```

I think this might be enough

src/uri/uri.cc
```diff
@@ -240,13 +244,39 @@ auto URI::path() const -> std::optional<std::string> {
...
+    return *this;
+  }
+
+  const auto is_relative_path = path.starts_with(".");
+  if (is_relative_path) {
+    throw URIError{"You cannot set a relative path"};
+  }
+
+  auto parsed_path = URI{std::move(path)};
+  this->path_ = parsed_path.path().value();
```
jviotti, line 278: Same here?

```suggestion
  this->path_ = URI{std::move(path)}.path_;
```

test/uri/uri_path_test.cc
```diff
@@ -1,64 +1,276 @@
...
+  EXPECT_EQ(uri.recompose(), "admin2");
+  EXPECT_EQ(path, "");
+}
+
+// TODO: dig why scheme return example.com
+// TEST(URI_path_setter_no_scheme, set_path_with_port) {
+//   sourcemeta::jsontoolkit::URI uri{"example.com:8080"};
+//   std::cout << "urn : " << uri.is_urn() << std::endl;
+//   std::cout << "tag : " << uri.is_tag() << std::endl;
+//   std::cout << "scheme : " << uri.scheme().value() << std::endl;
```
jviotti, line 272: Sounds like this whole TODO should be on `scheme()` instead? Here we don't really care, right? As `recompose()` will likely do the right thing anyway?

src/uri/uri.cc
tony-go: Move it where?

test/uri/uri_path_test.cc
tony-go: I want to test path here:

```
//   uri.path("/test");
//   EXPECT_EQ(uri.path().value(), "test");
```

tony-go: I found a very weird behavior:

```cc
TEST(URI_path_setter_no_scheme, set_path_with_port) {
  sourcemeta::jsontoolkit::URI uri{"example.com:8080"};

  uri.path("/test");
  EXPECT_EQ(uri.path().value(), "/test");
  EXPECT_EQ(uri.recompose(), "example.com:8080/test");
}
```

```
Expected equality of these values:
  uri.recompose()
    Which is: "example.com:///test"
  "example.com:8080/test"

[  FAILED  ] URI_path_setter_no_scheme.set_path_with_port (0 ms)
```

jviotti: Ah, alright. Interesting

jviotti (approved):

Issue #953, opened 2024-08-13
Establish proper test suite for `URI`


tony-go:
As much we move forward on #710, we figured out that writing a good tests suit is essential.

We want to have a test suit for RFC3986 URI https://www.rfc-editor.org/rfc/rfc3986! The goal of this issue is to gather every piece of information we could find on the topic.

tony-go:
RFC [2396](http://lists.w3.org/Archives/Public/www-qa/2001Aug/0000.html),

https://lists.w3.org/Archives/Public/www-qa/2001Aug/0000.html
https://www.w3.org/2004/04/uri-rel-test.html

tony-go:
What is your view on the finaly goal @jviotti ?

- Having a standalone test suite?
or
- Enrich our test suite with more cases?

tony-go:
I would like to share my vision about my previous comment.

I'd rather go for the second option, for two reasons:
1. We would like to benefit from our research as soon as can, by strengthen our test suite This is by doing a lot research that at some point we could have a fully compliant tests. (That could give, on the end, a standalone suite)
2. Even if having a fully compliant RFC3986 URI parser is the final goal, we should not forget that we first serve `jsontoolkit`. What I mean is that if we wait to have a full standalone suite before replacing `uriparser`, it could take a lot of time as once we finally have it we would like to make it review by fellow engineers...

In a one word my vision is to do a first research, extract all what we can from these, open PR (like [this one](https://github.com/sourcemeta/jsontoolkit/pull/957)) and then continue to do research on side while we move forward on critical topics for the project.

jviotti:
@tony-go Sorry if I wasn't clear at the beginning. I indeed meant enriching _our_ test suite with more tests, borrowing test suites from other places. The idea being that if we want to get rid of `uriparser`, let's have a more comprehensive set of tests (vs the basic ones I had right now) so we make sure we avoid pitfalls on complex areas like parsing or resolution.

jviotti:
Some other test suites I know:

- Hyperjump's URI parser: https://github.com/hyperjump-io/uri/blob/main/lib/parse.spec.ts, https://github.com/hyperjump-io/uri/blob/main/lib/resolve-reference.spec.ts, and https://github.com/hyperjump-io/uri/blob/main/lib/validation.spec.ts
[...]

PR #958, opened 2024-08-14
test(uri): add normalization tests according to RFC

tony-go:
Ref: #953

### DIFF
--- a/src/uri/uri.cc
+++ b/src/uri/uri.cc
@@ -316,6 +316,11 @@ auto URI::recompose_without_fragment() const -> std::optional<std::string> {
     }
   }
 
+  const auto user_info{this->userinfo()};
+  if (user_info.has_value()) {
+    result << user_info.value() << "@";
+  }
+
   // Host
   const auto result_host{this->host()};
   if (result_host.has_value()) {
--- a/test/uri/uri_recompose_test.cc
+++ b/test/uri/uri_recompose_test.cc
@@ -35,3 +35,88 @@ TEST(URI_recompose, empty_fragment) {
   const sourcemeta::jsontoolkit::URI uri{"#"};
   EXPECT_EQ(uri.recompose(), "#");
 }
+
+/*
+ * RFC 3986 normalization tests
+ */
+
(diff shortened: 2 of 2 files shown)


src/uri/uri.cc
```diff
@@ -316,6 +316,11 @@ auto URI::recompose_without_fragment() const -> std::optional<std::string> {
     }
   }
 
+  const auto user_info{this->userinfo()};
+  if (user_info.has_value()) {
+    result << user_info.value() << "@";
+  }
```
tony-go, line 322: One bug spotted! 🎉

tony-go:
I'll add more in a subsequent PR!

jviotti (approved):
Super cool!

test/uri/uri_recompose_test.cc
```diff
@@ -35,3 +35,88 @@ TEST(URI_recompose, empty_fragment) {
...
   EXPECT_EQ(uri.recompose(), "#");
 }
+
+/*
+ * RFC 3986 normalization tests
+ */
+
+// Inspired from https://cr.openjdk.org/~dfuchs/writeups/updating-uri/
+
+TEST(URI_normalize, rfc3986_1) {
```
jviotti, line 45: Actually, can we call these `URI_recompose`, as the test is really exercising normalisation _through_ recomposing?

tony-go: I am not sure tho! haha!

For me what we are testing is **normalisation** even if we test it through the recompose API. By the way I think that it is a good way to differentiate theses tests from what we have before.

WDYT?

tony-go: This is a preference tho! If it hurts you Im a okay to rename it :)

jviotti: So then let's have a `uri_normalize_test.cc` file with these? The current convention is that the test category matches the module + file, so a new file could work well?

tony-go: sold! I started with this apporach btw haha :)

jviotti: Haha awesome

PR #981, opened 2024-08-16
test(uri): add more normalize tests

tony-go:
Ref: #953

### DIFF
--- a/src/uri/uri.cc
+++ b/src/uri/uri.cc
@@ -236,7 +236,13 @@ auto URI::host() const -> std::optional<std::string_view> {
 auto URI::port() const -> std::optional<std::uint32_t> { return this->port_; }
 
 auto URI::path() const -> std::optional<std::string> {
+  // NOTE: This is a workaround for the fact that `uriparser` does not
+  // parse /.. as a segment, then we store nothing in the path_ field.
+  // By that we can't add the initial slash to the URI.
   if (!this->path_.has_value()) {
+    if (this->data == "/..") {
+      return "/";
+    }
     return std::nullopt;
   }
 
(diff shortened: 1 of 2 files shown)


src/uri/uri.cc
```diff
@@ -236,7 +236,13 @@ auto URI::host() const -> std::optional<std::string_view> {
 auto URI::port() const -> std::optional<std::uint32_t> { return this->port_; }
 
 auto URI::path() const -> std::optional<std::string> {
+  // NOTE: This is a workaround for the fact that `uriparser` does not
+  // parse /.. as a segment, then we store nothing in the path_ field.
+  // By that we can't add the initial slash to the URI.
   if (!this->path_.has_value()) {
+    if (this->data == "/..") {
+      return "/";
+    }
```
tony-go, line 245: cc @jviotti 

I was not really sure about how could I handle it :/

The culprit:

```cc
TEST(URI_normalize, path_4) {
  const sourcemeta::jsontoolkit::URI uri{"/.."};
  EXPECT_EQ(uri.recompose(), "/");
}
```

jviotti: Yeah, let's do anything that gets us to the tests passing for now. Once we are confident that our tests are strong and we start removing `uriparser`, we will then have the flexibility of fixing it properly. I imagine that once we have our own parser, etc and everything passes, then we'll have all of the puzzle pieces in front of us to do a proper refactoring effort.

jviotti (approved):

src/uri/uri.cc
tony-go: yeah :) 

I think that this is mostly because we are a wrapper around `uriparser` and that we don't really "handle" the segments.

Let see

PR #989, opened 2024-08-17
Avoid `BENCHMARK_MAIN` as we are already linking to `benchmark::benchmark_main`

jviotti:


jviotti:
cc @tony-go. I just realised this

PR #991, opened 2024-08-19
test(uri): add more parsing tests

tony-go:
I extracted a few tests from `uriparser`

### DIFF
--- a/src/uri/include/sourcemeta/jsontoolkit/uri.h
+++ b/src/uri/include/sourcemeta/jsontoolkit/uri.h
@@ -93,7 +93,7 @@ public:
   /// ```
   auto is_tag() const -> bool;
 
-  /// Check if the URI has `mailto` scheme. For example:
+  /// Check if the URI has the `mailto` scheme. For example:
   ///
   /// ```cpp
   /// #include <sourcemeta/jsontoolkit/uri.h>
--- a/src/uri/include/sourcemeta/jsontoolkit/uri.h
+++ b/src/uri/include/sourcemeta/jsontoolkit/uri.h
@@ -126,6 +126,17 @@ public:
   /// ```
   auto is_relative() const -> bool;
 
+  /// Check if the host is an ipv6 address. For example:
+  ///
+  /// ```cpp
+  /// #include <sourcemeta/jsontoolkit/uri.h>
+  /// #include <cassert>
+  ///
+  /// sourcemeta::jsontoolkit::URI uri{"http://[::1]"};
+  /// assert(uri.is_ipv6());
+  /// ```
+  auto is_ipv6() const -> bool;
+
   /// Get the scheme part of the URI, if any. For example:
   ///
   /// ```cpp
@@ -390,6 +401,7 @@ private:
   std::optional<std::string> scheme_;
   std::optional<std::string> fragment_;
   std::optional<std::string> query_;
+  bool is_ipv6_ = false;
 
   // Use PIMPL idiom to hide `urlparser`
   struct Internal;
--- a/src/uri/include/sourcemeta/jsontoolkit/uri.h
+++ b/src/uri/include/sourcemeta/jsontoolkit/uri.h
@@ -93,6 +93,17 @@ public:
   /// ```
   auto is_tag() const -> bool;
 
+  /// Check if the URI has `mailto` scheme. For example:
+  ///
+  /// ```cpp
+  /// #include <sourcemeta/jsontoolkit/uri.h>
+  /// #include <cassert>
+  ///
+  /// const sourcemeta::jsontoolkit::URI uri{"mailto:joe@example.com"};
+  /// assert(uri.is_mailto());
+  /// ```
+  auto is_mailto() const -> bool;
+
   /// Check if the URI only consists of a fragment. For example:
   ///
   /// ```cpp
--- a/src/uri/uri.cc
+++ b/src/uri/uri.cc
@@ -175,6 +175,7 @@ auto URI::parse() -> void {
   this->scheme_ = uri_text_range(&this->internal->uri.scheme);
   this->userinfo_ = uri_text_range(&this->internal->uri.userInfo);
   this->host_ = uri_text_range(&this->internal->uri.hostText);
+  this->is_ipv6_ = this->internal->uri.hostData.ip6 != nullptr;
   this->fragment_ = uri_text_range(&this->internal->uri.fragment);
   this->query_ = uri_text_range(&this->internal->uri.query);
   const auto port_text{uri_text_range(&this->internal->uri.portText)};
@@ -238,6 +239,8 @@ auto URI::is_mailto() const -> bool {
   const auto scheme{this->scheme()};
   return scheme.has_value() && scheme.value() == "mailto";
 }
+
+auto URI::is_ipv6() const -> bool { return this->is_ipv6_; }
 
 auto URI::is_fragment_only() const -> bool {
   return !this->scheme().has_value() && !this->host().has_value() &&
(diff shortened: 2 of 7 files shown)


test/uri/uri_parse_test.cc
```diff
@@ -0,0 +1,117 @@
...
+  EXPECT_THROW(
+      sourcemeta::jsontoolkit::URI uri{"http://[vA.123456"}, // missing "]"
+      sourcemeta::jsontoolkit::URIParseError);
+}
+
+// Inspired from
+// https://github.com/uriparser/uriparser/blob/bf0174e83164a4659c51c135399478bec389eafa/test/test.cpp#L315
+
+TEST(URI_parse, success_1) {
+  EXPECT_NO_THROW(sourcemeta::jsontoolkit::URI uri{
```
jviotti, line 38: Assuming this is parse test, should we assert on the results? At least by recomposing? Otherwise how do we know we are indeed parsing correctly (apart from not throwing?)

jviotti (approved):

test/uri/uri_parse_test.cc
tony-go: Yeah I try that :)

I face a few surprises!

jviotti: haha good surprises then

test/uri/uri_parse_test.cc
```diff
@@ -0,0 +1,173 @@
...
+  EXPECT_NO_THROW({
+    sourcemeta::jsontoolkit::URI uri{"../../"};
+    EXPECT_EQ(uri.recompose(), "../../");
+  });
+}
+
+TEST(URI_parse, root_path) {
+  EXPECT_NO_THROW({
+    sourcemeta::jsontoolkit::URI uri{"/"};
+    EXPECT_EQ(uri.recompose(), "");
```
tony-go, line 149: Always doubtful on this one :/

src/uri/include/sourcemeta/jsontoolkit/uri.h
```diff
@@ -93,6 +93,17 @@ class SOURCEMETA_JSONTOOLKIT_URI_EXPORT URI {
   /// ```
   auto is_tag() const -> bool;
 
+  /// Check if the URI has `mailto` scheme. For example:
```
jviotti, line 96:
```suggestion
  /// Check if the URI has the `mailto` scheme. For example:
```

test/uri/uri_parse_test.cc
```diff
@@ -0,0 +1,173 @@
...
+  EXPECT_THROW(
+      sourcemeta::jsontoolkit::URI uri{"http://[vA.123456"}, // missing "]"
+      sourcemeta::jsontoolkit::URIParseError);
+}
+
+// Inspired from
+// https://github.com/uriparser/uriparser/blob/bf0174e83164a4659c51c135399478bec389eafa/test/test.cpp#L315
+
+TEST(URI_parse, success_1) {
+  EXPECT_NO_THROW({
```
jviotti, line 38: I don't think you need these `EXPECT_NO_THROW` wrappers. Just omitting that and writing the test bodies implies that they will not throw.

test/uri/uri_parse_test.cc
```diff
@@ -0,0 +1,173 @@
...
+TEST(URI_parse, relative_2) {
+  EXPECT_NO_THROW({
+    sourcemeta::jsontoolkit::URI uri{"/one/two/three"};
+    EXPECT_EQ(uri.recompose(), "one/two/three");
+  });
+}
+
+TEST(URI_parse, relative_3) {
+  EXPECT_NO_THROW(
+      sourcemeta::jsontoolkit::URI uri{"//user:pass@localhost/one/two/three"});
```
jviotti, line 111: Should we recompose this one?

test/uri/uri_parse_test.cc
jviotti: I don't remember very well, but on a previous PR, I think the conclusion was that `""` and `"/"` are different URIs, as `https://google.com` and `https://google.com/` are indeed different URIs

test/uri/uri_parse_test.cc
```diff
@@ -0,0 +1,173 @@
...
+
+// Inspired from
+// https://github.com/uriparser/uriparser/blob/bf0174e83164a4659c51c135399478bec389eafa/test/test.cpp#L315
+
+TEST(URI_parse, success_1) {
+  EXPECT_NO_THROW({
+    sourcemeta::jsontoolkit::URI uri{
+        "//user:pass@[::1]:80/segment/index.html?query#frag"};
+    EXPECT_EQ(uri.recompose(),
+              "user:pass@::1:80/segment/index.html?query#frag");
```
jviotti, line 42: So the brackets are optional?

test/uri/uri_parse_test.cc
tony-go: fun story: it fails hahahahahahah

test/uri/uri_parse_test.cc
tony-go: I'll redo a test with `uriparser` itself, but it seems that this is how they parse it!

I'll check the spec too and probably write a few notes that could help us when we'll re-write the parser.

test/uri/uri_parse_test.cc
tony-go: I took a deeper look and I found interesting facts:

- `uriparser` test this case considering `"/"` as a [non empty segment](https://github.com/uriparser/uriparser/blob/bf0174e83164a4659c51c135399478bec389eafa/test/test.cpp#L121) that makes sens right? If there is a slash, there is an empty segment (at least). They use `pathTail` to check it. 
- When I parse `"/"` myself and that I explore our `internal` object I found this:
   ```text
   (lldb) p this->internal->uri
    (UriUriA) {
      scheme = (first = 0x0000000000000000, afterLast = 0x0000000000000000)
      userInfo = (first = 0x0000000000000000, afterLast = 0x0000000000000000)
      hostText = (first = 0x0000000000000000, afterLast = 0x0000000000000000)
      hostData = {
        ip4 = nullptr
        ip6 = nullptr
        ipFuture = (first = 0x0000000000000000, afterLast = 0x0000000000000000)
      }
      portText = (first = 0x0000000000000000, afterLast = 0x0000000000000000)
      pathHead = nullptr
      pathTail = nullptr
      query = (first = 0x0000000000000000, afterLast = 0x0000000000000000)
      fragment = (first = 0x0000000000000000, afterLast = 0x0000000000000000)
      absolutePath = 1
      owner = 1
      reserved = 0x0000000000000000
    }
   ``` 
   As you probably see, `pathTail` is `nullptr` hahaha!
   
 Two divergents facts! I think that if I figure out why I don't have this empty segment, I resolve the issue (more or less)

tony-go: Maybe a clue!

They use [`uriParseUriA`](https://github.com/uriparser/uriparser/blob/bf0174e83164a4659c51c135399478bec389eafa/test/test.cpp#L75) while we are using `uriParseSingleUriA`

jviotti: Ah, maybe. I'm not sure to be honest. At this point you likely know more about `uriparser` than I do. Keep up the good research and happy to pair debug if it helps

test/uri/uri_parse_test.cc
```diff
@@ -0,0 +1,139 @@
...
+
+TEST(URI_parse, relative_4) {
+  sourcemeta::jsontoolkit::URI uri{"../../"};
+  EXPECT_EQ(uri.recompose(), "../../");
+}
+
+TEST(URI_parse, root_path) {
+  sourcemeta::jsontoolkit::URI uri{"/"};
+  EXPECT_EQ(uri.recompose(), "/");
+}
```
tony-go, line 123: Fro this case, I had no choice other than check the original data (this.data), the good thing is that it helped me to skip a whole parsing job :)

src/uri/uri.cc
```diff
@@ -330,7 +358,16 @@ auto URI::recompose_without_fragment() const -> std::optional<std::string> {
   // Host
   const auto result_host{this->host()};
   if (result_host.has_value()) {
-    result << result_host.value();
+    if (this->is_ipv6()) {
+      // By default uriparser will parse the IPv6 address without brackets
+      // so we need to add them manually, as said in the RFC 2732:
+      // "To use a literal IPv6 address in a URL, the literal address should be
+      // enclosed in "[" and "]" characters."
+      // See https://tools.ietf.org/html/rfc2732#section-2
```
tony-go, line 366: The other one :)

test/uri/uri_parse_test.cc
tony-go: Bad that we did not have benchmark before :/ we could have seen an improvement.

jviotti: Haha nice! Yeah, we need more benchmarks. Overall a better process for evaluating performance and making sure we don't regress

jviotti (approved):

PR #1037, opened 2024-08-28
Don't emit evaluation callbacks if the callback is a no-op

jviotti:

### DIFF
--- a/src/jsonschema/compile_evaluate.cc
+++ b/src/jsonschema/compile_evaluate.cc
@@ -264,34 +264,28 @@ private:
   TargetType target_type_ = TargetType::Value;
 };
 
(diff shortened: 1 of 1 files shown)


jviotti:
On my machine, before:

```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
Draft4_Meta_1_No_Callback      0.025 ms        0.025 ms        28219
```

After:

```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
Draft4_Meta_1_No_Callback      0.015 ms        0.015 ms        47127
```

So a 40% improvement?

PR #1038, opened 2024-08-28
Create a single Google Benchmark executable

jviotti:
This will make it more practical to adopt
https://github.com/benchmark-action/github-action-benchmark. We can
still focus on specific subcomponents by either omitting the ones we
don't want at the CMake options level, or with `--benchmark_filter`.

### DIFF
--- a/benchmark/jsonschema_compiler.cc
+++ b/benchmark/jsonschema_compiler.cc
@@ -3,7 +3,7 @@
 #include <sourcemeta/jsontoolkit/json.h>
 #include <sourcemeta/jsontoolkit/jsonschema.h>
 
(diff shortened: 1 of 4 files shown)


benchmark/jsonschema_compiler.cc
```diff
@@ -22,4 +22,4 @@ static void Compiler_Basic(benchmark::State &state) {
   }
 }
 
-BENCHMARK(Compiler_Basic);
+BENCHMARK(JSONSchema_Compile_Basic);
```
jviotti, line 25: These more verbose prefixes make it easier to filter with `--benchmark_filter`

tony-go (approved):

PR #1043, opened 2024-08-28
Introduce basic tracing facilities for schema evaluation

jviotti:


jviotti:
With this, debug builds now emit signposts for the evaluator than can be read on Instruments. i.e.:

Screenshot 2024-08-28 at 3 57 56 PM

PR #1049, opened 2024-08-29
Give different platforms different benchmark names

jviotti:
Otherwise we seem to only get one comment on PRs for the run that
finished last.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,7 @@ jobs:
       - uses: benchmark-action/github-action-benchmark@v1
         if: matrix.platform.benchmark
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -143,6 +143,7 @@ jobs:
       - uses: benchmark-action/github-action-benchmark@v1
         if: matrix.platform.benchmark
         with:
+          name: ${{ runner.os }} (${{ matrix.platform.cc }}/${{ matrix.platform.type }})
```
jviotti, line 146:
```suggestion
          name: Benchmark ${{ runner.os }} (${{ matrix.platform.cc }}/${{ matrix.platform.type }})
```

PR #1056, opened 2024-08-29
Optimize how we determine instance location on target resolution

jviotti:

### DIFF
--- a/src/jsonschema/compile_evaluate.cc
+++ b/src/jsonschema/compile_evaluate.cc
@@ -123,20 +123,6 @@ public:
     return this->instance_location_;
   }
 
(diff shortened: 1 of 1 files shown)


src/jsonschema/compile_evaluate.cc
```diff
@@ -826,7 +814,14 @@ auto evaluate_step(
     context.push(annotation);
     EVALUATE_CONDITION_GUARD("SchemaCompilerAnnotationEmit", annotation,
                              instance);
-    const auto current_instance_location{context.instance_location(annotation)};
+    assert(annotation.target.second.empty());
+
+    // TODO: Can we avoid a copy of the instance location here?
```
tony-go, line 819: By moving the `instance location`?

PR #1106, opened 2024-09-03
refactor: get rid of args for Pointer class

tony-go:


jviotti:
@tony-go You have a conflict here ^

PR #1108, opened 2024-09-03
Use a vector of references to store annotation values

jviotti:


jviotti:
We can ignore this for now, as Draft 4 is free of annotations

PR #1173, opened 2024-09-06
feat(jsonpointer): introduce `WeakPointer`

tony-go:
The current `Pointer` abstraction stores properties by making copies instead of keeping references. The new `WeakPointer` is designed to eliminate unnecessary copies by holding references to the properties instead, thereby improving efficiency.

### DIFF
--- a/src/jsonpointer/include/sourcemeta/jsontoolkit/jsonpointer.h
+++ b/src/jsonpointer/include/sourcemeta/jsontoolkit/jsonpointer.h
@@ -32,7 +32,7 @@ namespace sourcemeta::jsontoolkit {
 using Pointer = GenericPointer<JSON::String>;
 
 /// @ingroup jsonpointer
-using WeakPointer = GenericPointer<std::reference_wrapper<std::string>>;
+using WeakPointer = GenericPointer<std::reference_wrapper<const std::string>>;
 
 /// @ingroup jsonpointer
 /// A global constant instance of the empty JSON Pointer.
--- a/src/jsonpointer/include/sourcemeta/jsontoolkit/jsonpointer.h
+++ b/src/jsonpointer/include/sourcemeta/jsontoolkit/jsonpointer.h
@@ -11,9 +11,10 @@
 #include <sourcemeta/jsontoolkit/jsonpointer_walker.h>
 #include <sourcemeta/jsontoolkit/uri.h>
 
-#include <memory>  // std::allocator
-#include <ostream> // std::basic_ostream
-#include <string>  // std::basic_string
+#include <functional> // std::reference_wrapper
+#include <memory>     // std::allocator
+#include <ostream>    // std::basic_ostream
+#include <string>     // std::basic_string
 
 /// @defgroup jsonpointer JSON Pointer
 /// @brief An growing implementation of RFC 6901 JSON Pointer.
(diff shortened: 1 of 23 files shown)


src/jsonpointer/include/sourcemeta/jsontoolkit/jsonpointer_token.h
```diff
@@ -158,11 +162,16 @@ template <typename PropertyT> class GenericToken {
   ///
   /// sourcemeta::jsontoolkit::Pointer::Token token{"foo"};
   /// assert(token.is_property());
+  ///
```
jviotti, line 165: Accidental?

test/jsonpointer/jsonpointer_weakpointer_test.cc
```diff
@@ -0,0 +1,272 @@
...
+  EXPECT_TRUE(pointer.at(0).is_property());
+  EXPECT_EQ(pointer.at(0).to_property(), "foo");
+  EXPECT_TRUE(pointer.at(1).is_property());
+  EXPECT_EQ(pointer.at(1).to_property(), "bar");
+  EXPECT_TRUE(pointer.at(2).is_property());
+  EXPECT_EQ(pointer.at(2).to_property(), "baz");
+  EXPECT_TRUE(pointer.at(3).is_property());
+  EXPECT_EQ(pointer.at(3).to_property(), "qux");
+
+  // TODO: we should assert that other has been moved
```
tony-go, line 204: I don't think we are moving correctly here! As there is still stuff in `other`!

test/jsonpointer/jsonpointer_weakpointer_test.cc
```diff
@@ -0,0 +1,272 @@
+#include <gtest/gtest.h>
+#include <sourcemeta/jsontoolkit/jsonpointer.h>
+#include <type_traits>
+
+const std::string foo = "foo";
```
jviotti, line 5:
```suggestion
static const std::string foo = "foo";
```

Can you make these static so that they are not visible from other files from the same compilation unit?

test/jsonpointer/jsonpointer_weakpointer_test.cc
jviotti: Its fine if there are still things, but they shouldn't be the same things

src/jsonpointer/include/sourcemeta/jsontoolkit/jsonpointer.h
```diff
@@ -31,6 +31,9 @@ namespace sourcemeta::jsontoolkit {
 /// @ingroup jsonpointer
 using Pointer = GenericPointer<JSON::String>;
 
+/// @ingroup jsonpointer
+using WeakPointer = GenericPointer<std::reference_wrapper<const std::string>>;
```
jviotti, line 35: @tony-go Can also add `#include <functional> // std::reference_wrapper`?

tony-go: Funny that it did not break on linux haha

jviotti (approved):

test/jsonpointer/jsonpointer_weakpointer_test.cc
tony-go: I'll dig in another PR!

PR #1181, opened 2024-09-11
feat: use weak pointer in `EvaluationContext`

tony-go:


src/jsonschema/compile_evaluate.cc
```diff
@@ -17,9 +17,19 @@
...
+    } else {                                                                   \
+      empty_pointer.push_back(token.to_index());                               \
+    }                                                                          \
+  }                                                                            \
+  assert(empty_pointer.size() == weak_pointer.size());
+
 class EvaluationContext {
 public:
-  using Pointer = sourcemeta::jsontoolkit::Pointer;
+  using Pointer = sourcemeta::jsontoolkit::WeakPointer;
```
tony-go, line 32: The initial change!

src/jsonpointer/include/sourcemeta/jsontoolkit/jsonpointer.h
```diff
@@ -59,6 +63,10 @@ const Pointer empty_pointer;
 SOURCEMETA_JSONTOOLKIT_JSONPOINTER_EXPORT
 auto get(const JSON &document, const Pointer &pointer) -> const JSON &;
 
+// TODO: add doc
```
tony-go, line 66: I'll do this later, when we'll agree on the global approach :)

src/jsonpointer/include/sourcemeta/jsontoolkit/jsonpointer.h
```diff
@@ -233,6 +245,12 @@ auto stringify(const Pointer &pointer,
                std::basic_ostream<JSON::Char, JSON::CharTraits> &stream)
     -> void;
 
+// TODO: add doc
+SOURCEMETA_JSONTOOLKIT_JSONPOINTER_EXPORT
+auto stringify(const WeakPointer &pointer,
+               std::basic_ostream<JSON::Char, JSON::CharTraits> &stream)
+    -> void;
+
```
tony-go, line 253: I had linker issues trying to "templating" these methods. I could give it a new shot in a another PR.

jviotti: I like the current approach of adding proper non-templated overloads here. Then on the `.cc` file, they can share an internal templated helper. Maybe worth extracting these already into a separate PR?

src/jsonschema/compile_evaluate.cc
```diff
@@ -169,8 +179,16 @@ class EvaluationContext {
...
+    Pointer relative_schema_location;
+    CONVERT_WEAK_TO_POINTER(relative_schema_location,
+                            step.relative_schema_location);
+    this->evaluate_path_.push_back(relative_schema_location);
+
+    Pointer relative_instance_location;
+    CONVERT_WEAK_TO_POINTER(relative_instance_location,
+                            step.relative_instance_location);
+    this->instance_location_.push_back(relative_instance_location);
+
```
tony-go, line 191: Steps or (`SchemaCompilerTemplate`) seems to be coupled to the compiler itself, this why I opt for this conversion step.

tony-go: This could be the culprit for slowness!

src/jsonpointer/jsonpointer.cc
```diff
@@ -78,6 +81,15 @@ auto get(const JSON &document, const Pointer::Token &token) -> const JSON & {
   }
 }
 
+auto get(const JSON &document,
```
jviotti, line 84: All of these changes look great. Can you split them already?

src/jsonschema/compile_evaluate.cc
```diff
@@ -169,8 +179,16 @@ class EvaluationContext {
     assert(step.relative_instance_location.size() <= 1);
     this->frame_sizes.emplace_back(step.relative_schema_location.size(),
                                    step.relative_instance_location.size());
-    this->evaluate_path_.push_back(step.relative_schema_location);
-    this->instance_location_.push_back(step.relative_instance_location);
+    Pointer relative_schema_location;
+    CONVERT_WEAK_TO_POINTER(relative_schema_location,
```
jviotti, line 193: This feels weird. I guess you do it because the instructions have `Pointer`, and you cannot push that to a `WeakPointer` yet. The solution seems to add, for `WeakPointer`, a `push_back` overload that takes non-weak pointers?

src/jsonschema/compile_evaluate.cc
jviotti: Yep. See my other comment. I think what we need is a `WeakPointer::push_back` overload that takes `Pointer`

src/jsonschema/compile_evaluate.cc
```diff
@@ -17,9 +17,18 @@
 
 namespace {
 
+#define PUSH_BACK(weak_pointer, target)                                        \
```
jviotti, line 20:
```suggestion
#define PUSH_BACK(pointer, target)                                        \
```

The first argument is the non-weak one. In any case, hopefully we move this to `.push_back()` to get rid of the macro

test/jsonschema/jsonschema_test_utils.h
```diff
@@ -5,6 +5,22 @@
 #include <tuple>
 #include <vector>
 
+namespace {
+std::string flatten_pointer(const auto &pointer) {
```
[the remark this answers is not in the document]
tony-go: Yeah I did not realised!

I need it, because we compare WeakPointer and Pointer. I'll use stringify

PR #1197, opened 2024-09-16
refactor: clean bundle from std::future

tony-go:


tony-go:
last item: update the doc

tony-go:
It slows down on macOS... hmm maybe the move that I forgot

src/jsonschema/bundle.cc
```diff
@@ -209,20 +208,16 @@ auto bundle(sourcemeta::jsontoolkit::JSON &schema, const SchemaWalker &walker,
...
             const std::optional<std::string> &default_dialect)
-    -> std::future<sourcemeta::jsontoolkit::JSON> {
+    -> sourcemeta::jsontoolkit::JSON {
   sourcemeta::jsontoolkit::JSON copy = schema;
-  bundle(copy, walker, resolver, options, default_dialect).wait();
-  std::promise<sourcemeta::jsontoolkit::JSON> promise;
-  promise.set_value(std::move(copy));
-  return promise.get_future();
+  bundle(copy, walker, resolver, options, default_dialect);
+  return copy;
```
tony-go, line 220: I should avoid this copy

tony-go: Oh now I got it. This function is prolly a convenience for the user which don't have to do the copy

tony-go: but this is used in `compile`:

```cpp
auto compile(const JSON &schema, const SchemaWalker &walker,
             const SchemaResolver &resolver, const SchemaCompiler &compiler,
             const std::optional<std::string> &default_dialect)
    -> SchemaCompilerTemplate {
  assert(is_schema(schema));

  // Make sure the input schema is bundled, otherwise we won't be able to
  // resolve remote references here
  const JSON result{bundle(schema, walker, resolver, BundleOptions::Default,
                           default_dialect)};

```

tony-go: If we know that we'll modify the schema I ownder if we should not have

```
auto compile(JSON &schema, const SchemaWalker &walker,
             const SchemaResolver &resolver, const SchemaCompiler &compiler,
             const std::optional<std::string> &default_dialect)
    -> SchemaCompilerTemplate;
    ```

Comment (tony-go) [reply src/jsonschema/bundle.cc:220]: and then we could use direclty:

```
auto bundle(sourcemeta::jsontoolkit::JSON &schema, const SchemaWalker &walker,
            const SchemaResolver &resolver,
            const BundleOptions options = BundleOptions::Default,
            const std::optional<std::string> &default_dialect = std::nullopt)
    -> void;
```

jviotti: Yeah, maybe. For now I wouldn't worry very much about it, as we are much more focused on improving evaluation time than compilation.

jviotti (approved):

src/jsonschema/bundle.cc
tony-go: I mean we regress a bit https://github.com/sourcemeta/jsontoolkit/pull/1197#pullrequestreview-2306242698

jviotti: I don't see any difference there. Also, your changes shouldn't affect the validation benchmarks, as we don't take compilation into account? If you saw any difference, it might be just transient, as we could get some noise depending on the load of GitHub Action's runners, etc

tony-go: Oh right! because the first benchmark I ran highlighted a diff of more than 100ms this is why i was worried haha

PR #1205, opened 2024-09-18
chore: enable default compiler options with noa

tony-go:


tony-go:
It is worst globally haha

PR #1206, opened 2024-09-18
chore: enable compiler opti

tony-go:


tony-go:
I tried different flags: `-O3`, `-O2`, `-O1`. The fun fact is that the more optimized degrades more benchmark than less optimized ones.

tony-go:
and `-Ofast` make some test failures

PR #1210, opened 2024-09-23
refactor: use `std::uint8_t` for enum

tony-go:
First advice taken from the great tidy!

### DIFF
--- a/src/jsonschema/compile_evaluate.cc
+++ b/src/jsonschema/compile_evaluate.cc
@@ -1,4 +1,3 @@
-#include <cstdint>
 #include <sourcemeta/jsontoolkit/jsonschema.h>
 #include <sourcemeta/jsontoolkit/jsonschema_compile.h>
 #include <sourcemeta/jsontoolkit/uri.h>
@@ -7,6 +6,7 @@
 
 #include <algorithm>   // std::min, std::any_of
 #include <cassert>     // assert
+#include <cstdint>     // std::uint8_t
 #include <functional>  // std::reference_wrapper
 #include <iterator>    // std::distance, std::advance
 #include <limits>      // std::numeric_limits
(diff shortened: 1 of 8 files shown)


src/jsonschema/compile_evaluate.cc
```diff
@@ -1,3 +1,4 @@
+#include <cstdint>
```
jviotti, line 1: Can you move this to the standard library includes within line 8 to line 17?

src/jsonschema/include/sourcemeta/jsontoolkit/jsonschema_anchor.h
```diff
@@ -3,6 +3,7 @@
 
 #include "jsonschema_export.h"
 
+#include <cstdint>
```
jviotti, line 6: Also here, can you move this below to the std includes section?

src/jsonschema/compile_evaluate.cc
tony-go: HAHAHAH the formater added them automatically!

tony-go:
Let me re-add headers properly

src/jsonschema/compile_evaluate.cc
```diff
@@ -6,6 +6,7 @@
 
 #include <algorithm>   // std::min, std::any_of
 #include <cassert>     // assert
+#include <cstdint>     // std::uint64_t
```
jviotti, line 9:
```suggestion
#include <cstdint>     // std::uint8_t
```

?

jviotti (approved):
I made a tiny suggestion but other than that it looks good!

src/jsonschema/compile_evaluate.cc
tony-go: dear focus!

PR #1231, opened 2024-09-25
feat(json): expose JSON::try_at

tony-go:
The final goal is to replace [this code](https://github.com/sourcemeta/jsontoolkit/blob/main/src/jsonpointer/jsonpointer.cc#L69-L70), that do two check instead of one.


tony-go:
Error in one header, let me fix it.

src/json/include/sourcemeta/jsontoolkit/json_value.h
```diff
@@ -847,6 +848,36 @@ class SOURCEMETA_JSONTOOLKIT_JSON_EXPORT JSON {
...
+  /// ```cpp
+  /// #include <sourcemeta/jsontoolkit/json.h>
+  /// #include <cassert>
+  ///
+  /// const sourcemeta::jsontoolkit::JSON document =
+  ///   sourcemeta::jsontoolkit::parse("{ \"0\": 1 }");
+  /// const auto result = document.try_at(0);
+  /// EXPECT_TRUE(result.has_value());
+  /// EXPECT_EQ(result.value().to_integer(), 1);
+  [[nodiscard]] auto try_at(const typename Array::size_type index) const
```
tony-go, line 878: Not sure we need so far! We could prolly delete it!

src/json/json_value.cc
```diff
@@ -450,6 +450,23 @@ auto JSON::operator-=(const JSON &substractive) -> JSON & {
   }
 }
 
+[[nodiscard]] auto JSON::try_at(const JSON::String &key) const
+    -> std::optional<JSON> {
+  assert(this->is_object());
+  const auto value{std::get<Object>(this->data).data.find(key)};
+  if (value == std::get<Object>(this->data).data.end()) {
```
tony-go, line 457: I should avoid double get

src/json/json_value.cc
```diff
@@ -450,6 +450,19 @@ auto JSON::operator-=(const JSON &substractive) -> JSON & {
...
 }
 
+[[nodiscard]] auto JSON::try_at(const JSON::String &key) const
+    -> std::optional<JSON> {
+  assert(this->is_object());
+
+  const auto &object{std::get<Object>(this->data)};
+  const auto value{object.data.find(key)};
+
+  if (value == object.data.end()) {
```
jviotti, line 460:
```suggestion
  if (value == object.data.cend()) {
```

Just to make the intention clearer?

jviotti (approved):

src/json/json_value.cc
```diff
@@ -450,6 +450,19 @@ auto JSON::operator-=(const JSON &substractive) -> JSON & {
   }
 }
 
+[[nodiscard]] auto JSON::try_at(const JSON::String &key) const
+    -> std::optional<JSON> {
```
jviotti, line 454: @tony-go Actually, I think this interface is not great. The result will be a copy, whereas we want an optional _reference_ to `JSON`. I think you will need `std::reference_wrapper` around it though:

```suggestion
    -> std::optional<std::reference_wrapper<JSON>> {
```

src/jsonpointer/jsonpointer.cc
jviotti, line 1: I guess these changes are outside the scope of this PR?

tony-go: nope, we should rename it

PR #1241, opened 2024-09-26
Avoid double-hashing on `.try_get()`

tony-go:


src/json/include/sourcemeta/jsontoolkit/json_value.h
```diff
@@ -861,7 +861,8 @@ class SOURCEMETA_JSONTOOLKIT_JSON_EXPORT JSON {
   /// const auto result = document.try_at("foo");
   /// EXPECT_TRUE(result.has_value());
   /// EXPECT_EQ(result.value().to_integer(), 1);
```
jviotti, line 863:
```suggestion
  /// EXPECT_EQ(result.value().get().to_integer(), 1);
```

The example needs to be updated

src/jsonpointer/jsonpointer.cc
```diff
@@ -66,8 +66,9 @@ auto try_traverse(const sourcemeta::jsontoolkit::JSON &document,
     const auto &instance{current.get()};
     if (token.is_property()) {
       const auto &property{token.to_property()};
-      if (instance.defines(property)) {
-        current = instance.at(property);
+      const auto json_value{instance.try_at(property)};
```
jviotti, line 69:
```suggestion
      auto json_value{instance.try_at(property)};
```

src/jsonpointer/jsonpointer.cc
```diff
@@ -66,8 +66,9 @@ auto try_traverse(const sourcemeta::jsontoolkit::JSON &document,
...
+      if (json_value.has_value()) {
+        current = json_value.value();
```
jviotti, line 71:
```suggestion
        current = std::move(json_value.value());
```

jviotti (approved):

PR #1244, opened 2024-10-01
Use the `nosubs` option to `std::regex`

michaelmior:
This is possible since any matching groups will not be exposed in the final schema and might make matching slightly more efficient.


jviotti (approved):
Oh, awesome! I didn't know about this

jviotti:
Looks like the benchmark integration is failing to post the comments because you are an external collaborator. Let me just merge and try to fix it later.

michaelmior:
@jviotti I think there's probably a lot of interesting things on the regex side that we can optimize later. There is also `std::regex::optimize` which might make matching faster at the expense of possibly making building the regex slower.

I think what would be *really* interesting is to compile some regexes ahead of time as part of the schema compilation process. I think a lot of regexes for `patternProperties` can be reduced to string prefix checks for example.

PR #1250, opened 2024-10-03
benchmark: add nested oneof (draft4)

tony-go:


benchmark/schemas/draft7_nested_oneof.json
jviotti, line 1: I think this is too complicated. It's exercising many other keywords apart from `oneOf`. Can we have an artificial concise schema that is only `oneOf` plus potentially a few `type`? We want the schema to be by far mostly about the thing we are benchmarking, whereas here the other keywords will introduce noise.

I think it will also end up being small enough to inline on the benchmark file too

jviotti: Also, can we make it Draft 4? `oneOf` in Draft 7 is the same as Draft 4, so by having the benchmark on Draft 4, we cover it in all of them

tony-go: Okay :)

jviotti (approved):

PR #1272, opened 2024-10-05
Emit loop vectorization diagnostics on the schema evaluator

jviotti:


jviotti:
Screenshot 2024-10-05 at 5 39 51 pm

PR #1360, opened 2024-12-01
Fix string overrun in URL canonicalization

michaelmior:
This is minor, the way URL canonicalization is currently written, it's possible that the implementation will access a character past the end of the string.


jviotti (approved):
Thanks!

PR #1581, opened 2025-02-25
Add the first 31 bytes to the hash of long strings

michaelmior:
This adds the first 31 bytes of the string into the hash value when we are dealing with longer strings.
This should serve to reduce hash collisions and consequently, also eliminate some string comparisons.


michaelmior:
Note that hashing long strings is significantly slower. This is expected, but I don't think it's a problem. We're sacrificing a bit more time to hash the strings to avoid collisions in the future. Since collisions always mean a string comparison of longer strings, I think this will be a significant win overall. We may want another microbenchmark to capture this.

jviotti:
Let me open a Blaze PR with this so we can see the GoogleBenchmark impact and also on the final benchmark

jviotti:
See https://github.com/sourcemeta/blaze/pull/381

jviotti:
Closing this as I think we confirmed that there was not much of a boost after all. We can revisit if needed

michaelmior:
I think it could still be worthwhile since it should only add a tiny amount of overhead. There was one test case that had long strings that I think benefited here.

jviotti:
We are getting some conflicts though. Would you want to fix them, given you created this PR?

michaelmior:
@jviotti Take a look as I think this should be fixed now.

jviotti (approved):
Looks good to me! Do you want me to open a branch to test this in Blaze first or are we fine with merging? I think we did run the benchmark with this change and it was good, right?

michaelmior:
@jviotti Merging sounds good to me. Although it couldn't hurt to run the benchmark again first.

jviotti:
Here is a Blaze PR with it: https://github.com/sourcemeta/blaze/pull/480

jviotti:
@michaelmior The Blaze PR is failing as this branch is not up to date. Do you mind rebasing to `main` or merging `main` back in?

michaelmior:
@jviotti Rebased!

jviotti:
Here are the results: https://github.com/sourcemeta-research/jsonschema-benchmark/actions/runs/16473101584. Looks great to me. Let's merge!

PR #1640, opened 2025-05-12
Run ClangTidy on GitHub Actions

jviotti:
See: https://github.com/sourcemeta/blaze/issues/429

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,3 +154,18 @@ jobs:
           alert-threshold: '5%'
           comment-always: true
           fail-on-alert: false
+
+  clang-tidy:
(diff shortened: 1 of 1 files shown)


bavulapati:
can we rerun this on the latest main commit?

jviotti:
Rebased! The results will show up in a bit

jviotti:
For later (I don't want to block this PR even more), we can make GitHub Actions always install the latest LLVM version using the script in here: https://apt.llvm.org

jviotti:
I think we were relying on `ClangTidy` being available by default on the runner. Maybe they removed it now? Probably a good idea to install it

jviotti:
Just rebased this PR to include your changes to download the ClangTidy binary

PR #1647, opened 2025-05-18
Include clean src core jsonl

bavulapati:
Refs: https://github.com/sourcemeta/blaze/issues/429

### DIFF
--- a/src/core/jsonl/include/sourcemeta/core/jsonl.h
+++ b/src/core/jsonl/include/sourcemeta/core/jsonl.h
@@ -1,11 +1,11 @@
 #ifndef SOURCEMETA_CORE_JSONL_H_
 #define SOURCEMETA_CORE_JSONL_H_
 
+#include "sourcemeta/core/json_value.h"
 #ifndef SOURCEMETA_CORE_JSONL_EXPORT
 #include <sourcemeta/core/jsonl_export.h>
 #endif
 
-#include <sourcemeta/core/json.h>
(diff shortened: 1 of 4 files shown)


src/core/jsonl/include/sourcemeta/core/jsonl.h
```diff
@@ -1,11 +1,11 @@
 #ifndef SOURCEMETA_CORE_JSONL_H_
 #define SOURCEMETA_CORE_JSONL_H_
 
+#include "sourcemeta/core/json_value.h"
```
jviotti, line 4: Same as in the other PR. Use `<...>` includes and please keep the ordering. This should be close to the line you removed and not as a standalone thing here

src/core/jsonl/iterator.cc
```diff
@@ -1,7 +1,10 @@
+#include <istream>
```
jviotti, line 1: Please put this with the other system includes and add a comment to why we are requiring it

src/core/jsonl/jsonl.cc
```diff
@@ -1,3 +1,5 @@
+#include "sourcemeta/core/json_value.h"
+#include <istream>
```
jviotti, line 2: Same here. Please put system includes AFTER sourcemeta includes with a blank line in between, and with a comment noting down why its being included.

bavulapati:
@jviotti Addressed review comments. PTAL

jviotti (approved):

PR #1648, opened 2025-05-18
style(clang-tidy): Clean include headers of src/core/jsonpointer

bavulapati:
Refs: sourcemeta/blaze#429

### DIFF
--- a/src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
+++ b/src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
@@ -1,14 +1,14 @@
 #ifndef SOURCEMETA_CORE_JSONPOINTER_H_
 #define SOURCEMETA_CORE_JSONPOINTER_H_
 
-#include "sourcemeta/core/json_hash.h"
-#include "sourcemeta/core/json_value.h"
 #ifndef SOURCEMETA_CORE_JSONPOINTER_EXPORT
 #include <sourcemeta/core/jsonpointer_export.h>
 #endif
 
 #include <sourcemeta/core/uri.h>
 
+#include <sourcemeta/core/json_hash.h>
+#include <sourcemeta/core/json_value.h>
 #include <sourcemeta/core/jsonpointer.h>
 #include <sourcemeta/core/jsonpointer_pointer.h>
 #include <sourcemeta/core/jsonpointer_subpointer_walker.h>
--- a/src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
+++ b/src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
@@ -1,17 +1,16 @@
 #ifndef SOURCEMETA_CORE_JSONPOINTER_H_
 #define SOURCEMETA_CORE_JSONPOINTER_H_
 
+#include "sourcemeta/core/json_hash.h"
+#include "sourcemeta/core/json_value.h"
 #ifndef SOURCEMETA_CORE_JSONPOINTER_EXPORT
 #include <sourcemeta/core/jsonpointer_export.h>
 #endif
 
-#include <sourcemeta/core/json.h>
 #include <sourcemeta/core/uri.h>
 
 #include <sourcemeta/core/jsonpointer.h>
-#include <sourcemeta/core/jsonpointer_error.h>
 #include <sourcemeta/core/jsonpointer_pointer.h>
-#include <sourcemeta/core/jsonpointer_position.h>
 #include <sourcemeta/core/jsonpointer_subpointer_walker.h>
 #include <sourcemeta/core/jsonpointer_template.h>
 #include <sourcemeta/core/jsonpointer_walker.h>
--- a/src/core/jsonpointer/include/sourcemeta/core/jsonpointer_pointer.h
+++ b/src/core/jsonpointer/include/sourcemeta/core/jsonpointer_pointer.h
@@ -1,11 +1,11 @@
 #ifndef SOURCEMETA_CORE_JSONPOINTER_POINTER_H_
 #define SOURCEMETA_CORE_JSONPOINTER_POINTER_H_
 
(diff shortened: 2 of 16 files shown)


test/jsonpointer/jsonpointer_error_test.cc
```diff
@@ -1,6 +1,7 @@
 #include <gtest/gtest.h>
 #include <sourcemeta/core/json.h>
 #include <sourcemeta/core/jsonpointer.h>
+#include <sourcemeta/core/jsonpointer_error.h>
```
bavulapati, line 4: This makes the includes better but should the consumer of `jsonpointer` be aware of more than one header file?

jviotti: I think it's fine for test files!

src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
```diff
@@ -1,17 +1,16 @@
 #ifndef SOURCEMETA_CORE_JSONPOINTER_H_
 #define SOURCEMETA_CORE_JSONPOINTER_H_
 
+#include "sourcemeta/core/json_hash.h"
```
jviotti, line 4: Can you avoid `"..."` includes in favour of `<...>` includes?

jviotti: Also, there is an ordering of includes in most of these files. You can notice that the Sourcemeta includes are below

src/core/jsonpointer/include/sourcemeta/core/jsonpointer_pointer.h
```diff
@@ -1,14 +1,14 @@
 #ifndef SOURCEMETA_CORE_JSONPOINTER_POINTER_H_
 #define SOURCEMETA_CORE_JSONPOINTER_POINTER_H_
 
+#include <cstddef>
```
jviotti, line 4: Most headers follow a specific header ordering: the Sourcemeta headers followed by the standard include headers. Can you respect that ordering? You can put `cstddef` below, and also as a comment put why you are including it.

jviotti:
Looks good but please keep in mind header include ordering, using `<...>` instead of `"..."` and for system headers, put a comment next to it marking which symbols you are including for each (this makes it easier to remember why I'm pulling them)

src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
bavulapati: > Can you avoid `"..."` includes in favour of `<...>` includes?

Why is `<... >` preferred for non C++ standard headers?

src/core/jsonpointer/include/sourcemeta/core/jsonpointer_pointer.h
bavulapati: Is there a reason for this order?

jviotti: Just convention. Otherwise the headers are all over the place

src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
jviotti: `"..."` depends on position, while `<...>` depends on include paths. With the latter, it is easier to control for more advanced setups

bavulapati:
@jviotti addressed the comments. PTAL

src/core/jsonpointer/include/sourcemeta/core/jsonpointer_position.h
```diff
@@ -1,11 +1,13 @@
 #ifndef SOURCEMETA_CORE_JSONPOINTER_POSITION_H_
 #define SOURCEMETA_CORE_JSONPOINTER_POSITION_H_
 
+#include "sourcemeta/core/json_hash.h"
```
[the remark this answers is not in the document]
bavulapati: Thanks for catching it. 👍🏾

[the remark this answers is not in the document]
bavulapati: @jviotti fixed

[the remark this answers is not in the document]
bavulapati: I'm thinking about C++ modules. I'm learning more about them. 
Do you have any thoughts about using modules?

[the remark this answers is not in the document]
jviotti: They are cool but not widely available / supported yet. It will likely be a few hours until they are common enough to be adopted on a project like this!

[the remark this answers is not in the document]
bavulapati: There are limitations about support documented [here](https://cmake.org/cmake/help/latest/manual/cmake-cxxmodules.7.html#compiler-support).

But if the limitations fit the project, it would be cool to use modules.
It would reduce the compile time, avoid header files, improve readability etc.,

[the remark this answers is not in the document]
jviotti: I think it is too early. There are other companies using these projects so the limitations extend beyond just me. I will probably look into it only once it is mainstream and supported in pretty much all compiler versions in use in i.e. major Linux distros, etc.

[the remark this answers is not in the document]
bavulapati: The fact that modules are half a decade old, surprises me! 🤷🏾‍♂️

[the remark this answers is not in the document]
jviotti: Well, we are on C++20, which is technically 5 years old, as I found it tricky to use C++23 still. Default GitHub Actions runners won't handle it.

PR #1650, opened 2025-05-21
style(clang-tidy): Clean include headers of src/core/jsonschema

bavulapati:
Refs: sourcemeta/blaze#429


bavulapati:
These changes are getting bigger when I consider one folder at a time

src/core/jsonschema/bundle.cc
```diff
@@ -1,8 +1,16 @@
...
-#include <cassert> // assert
-#include <sstream> // std::ostringstream
-#include <utility> // std::move
+#include <sourcemeta/core/jsonschema_error.h>
+#include <sourcemeta/core/jsonschema_frame.h>
+#include <sourcemeta/core/jsonschema_types.h>
+
+#include <cassert>  // assert
+#include <cstddef>  // std::size_t
+#include <optional> //std::nullopt
```
jviotti, line 10:
```suggestion
#include <optional> // std::nullopt
```

src/core/jsonschema/official_resolver.in.cc
jviotti, line 1: This file needs to have its headers ordered properly

test/jsonschema/jsonschema_bundle_2019_09_test.cc
jviotti, line 1: This feels weird. Why do we need to include `jsonschema_error.h` if we are already including `jsonschema.h`, which includes that other one?

bavulapati: [SchemaResolutionError](https://github.com/bavulapati/core/blob/e6f0573627a639b74eefea718cd347bd32efd0be/src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h#L39-L40) 

```
/home/personal/repos/sourcemeta/core/test/jsonschema/jsonschema_bundle_2019_09_test.cc:232:25: error: ‘SchemaResolutionError’ in namespace ‘sourcemeta::core’ does not name a type
  232 |       sourcemeta::core::SchemaResolutionError);
```

bavulapati:
@jviotti PTAL

test/jsonschema/jsonschema_bundle_2019_09_test.cc
jviotti: Is that a compilation error or a ClangTidy error? Because you can clearly see that `jsonschema.h` includes `jsonschema_error.h` here: https://github.com/sourcemeta/core/blob/main/src/core/jsonschema/include/sourcemeta/core/jsonschema.h#L11.

And of course things have been working so far. Otherwise with this change we start making external consumers (like tests) start requiring private internal headers?

jviotti: Looking at some other changes, I think it wants you to _directly_ include every header that's required, without considering transitive includes. In that case, can you disable the rule in the specific parts instead? i.e. if we include a public header, add a `NOLINT` comment for this specific rule? Or maybe there is a way to configure ClangTidy to understand this. This is the config file: https://github.com/sourcemeta/core/blob/main/cmake/common/targets/clang-tidy.config

jviotti: OK, I found the problem, and fixed it in https://github.com/sourcemeta/core/pull/1651 alongside some other minor things.

The thing is that ClangTidy led is in the wrong direction in some cases. For example, it made you remove the `json_error.h` include on `json.h`, claiming that `json.h` doesn't use it. Which is correct, as it only includes it to make it available for others.

I fixed it now by re-adding some includes, but what we need to do is not let ClangTidy warn about removal of private header includes on a public header, which `NOLINT` comments, etc.

src/core/jsonschema/include/sourcemeta/core/jsonschema.h
```diff
@@ -5,20 +5,15 @@
 #include <sourcemeta/core/jsonschema_export.h>
 #endif
 
-#include <sourcemeta/core/json.h>
+#include <sourcemeta/core/json.h> // NOLINT(misc-include-cleaner) Public header needs it
```
bavulapati, line 8: @jviotti Let me know if this is not what you meant

test/jsonschema/jsonschema_bundle_2019_09_test.cc
bavulapati: But How do we know if a header needs to be excluded from Lint?

src/core/jsonschema/include/sourcemeta/core/jsonschema.h
jviotti: I mean that for the public header of each module, we ignore its private includes. For example, in `jsonschema.h`, I presume you can do this:

```c++
// NOLINTBEGIN(misc-include-cleaner)

#include <sourcemeta/core/jsonschema_error.h>
#include <sourcemeta/core/jsonschema_frame.h>
#include <sourcemeta/core/jsonschema_resolver.h>
#include <sourcemeta/core/jsonschema_transform.h>
#include <sourcemeta/core/jsonschema_types.h>
#include <sourcemeta/core/jsonschema_walker.h>

// NOLINTEND(misc-include-cleaner)
```

Essentially, for a header `X.h`, then don't clean its `X_<private>.h` includes

jviotti: And we can do the same for every public header of every module

bavulapati: Sure for every public header of each module.

bavulapati:
@jviotti addressed it. PTAL
I'll send a separate PR for all the other modules

bavulapati:
@jviotti Consider a public header like `json.h`, it includes a bunch of private headers like `json_error.h`, `json_value.h`. 

What do you think about making `json.h` complete, having all the declarations of it's private headers?
In modules, there is a concept of re-export. If a module B is imported into module A, module A can re export the imported module A.
That would make the consumer(dev) to just have the knowledge of a single header, instead of going through chain of modules?

bavulapati:
In our case of clang-tidy, even though the public header `json.h` is included, the check `misc-include-cleaner` expects `json_error.h` to be included, if using a declaration from `json-error.h`

jviotti:
> What do you think about making json.h complete, having all the declarations of it's private headers?
In modules, there is a concept of re-export. If a module B is imported into module A, module A can re export the imported module A.

The idea of these private headers was to make things a little bit more maintainable. Otherwise these modules often expose a lot of stuff, and the single header file would get huge

test/jsonschema/jsonschema_base_dialect_test.cc
jviotti, line 1: These tests look off. The tests should not be including `jsonschema_error` if they already include `jsonschema`, as that one includes the other?

test/jsonschema/jsonschema_flat_file_resolver_test.cc
```diff
@@ -1,6 +1,8 @@
 #include <gtest/gtest.h>
 
 #include <sourcemeta/core/jsonschema.h>
+#include <sourcemeta/core/jsonschema_error.h>
```
jviotti, line 4: This one looks off too?

test/jsonschema/jsonschema_official_walker_2019_09_test.cc
```diff
@@ -1,6 +1,8 @@
 #include <gtest/gtest.h>
 
+#include <sourcemeta/core/json.h>
 #include <sourcemeta/core/jsonschema.h>
+#include <sourcemeta/core/jsonschema_walker.h>
```
jviotti, line 5: These ones too, as `jsonschema.h` already includes `jsonschema_walker.h`

jviotti:
Or maybe we can turn of `MissingIncludes`: https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html

test/jsonschema/jsonschema_base_dialect_test.cc
bavulapati: As I mentioned earlier, clang-tidy expects the includes to be direct not transitive

bavulapati:
> Hmm, well that will be tricky then. This whole part of ClangTidy looks complicated. Let me do some more digging. Maybe we should start with other warnings first?

Yes, we can try that too.

bavulapati:
> See #1685. I think that is enough for now. Maybe worth trying to re-create that PR given the ClangTidy warnings with that new config?
> 
> I'll think more about a better solution long term. Maybe C++ modules fixes all of this.

Yeah, C++ modules definitely makes life easier. I thought of adding an issue or TODO to use modules in the future. But I'm wondering, How do we know that the time is right?

jviotti:
> How do we know that the time is right?

I think we'll know once most popular C++ libraries out there support it

jviotti:
Hey @bavulapati , I'm not sure that are you still working on these, so let me close them to reduce noise. Always feel free to re-open if you ever get back to it

PR #1687, opened 2025-05-24
style(clang-tidy): Use use nolint on public headers

bavulapati:
Ref: https://github.com/sourcemeta/core/pull/1650#discussion_r2104799191


jviotti (approved):
Awesome!

PR #1702, opened 2025-05-27
chore: move clang-tidy config file to project directory

bavulapati:
clang-tidy when the config is not explicitly mentioned in the command looks for it in the project directory with the name `.clang-tidy`. And editors like VS Code recognizes the `.clang-tidy` file when code analysis is enabled. This is not a fix for a bug. This is just trying to make clang-tidy config and usage more visible.


cmake/common/targets/clang-tidy.cmake
```diff
@@ -35,13 +35,11 @@ function(sourcemeta_target_clang_tidy)
   file(GLOB_RECURSE SOURCEMETA_TARGET_CLANG_TIDY_FILES
     ${SOURCEMETA_TARGET_CLANG_TIDY_SOURCES})
 
-  set(CLANG_TIDY_CONFIG "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/clang-tidy.config")
```
jviotti, line 38: Instead of removing this, can you tweak the variable to point at `${PROJECT_DIR}/.clang-tidy` (or whatever the right variable name is?)

.clang-tidy
jviotti, line 1: If we are moving this file, can we also do the same with the `clang-format.config` config file in the same directory?

bavulapati:
not needed

PR #1732, opened 2025-06-20
style(clang-tidy): fix bugprone-return-const-ref-from-parameter errors

### LINKED ISSUE #1657: ClangTidy warnings for `bugprone-return-const-ref-from-parameter`
```
/Users/jviotti/Projects/core/src/core/jsonpointer/jsonpointer.cc:142:12: error: returning a constant reference parameter may cause use-after-free when the parameter is constructed from a temporary [bugprone-return-const-ref-from-parameter,-warnings-as-errors]
  142 |     return document;
      |            ^~~~~~~~

/Users/jviotti/Projects/core/src/core/jsonpointer/jsonpointer.cc:150:12: error: returning a constant reference parameter may cause use-after-free when the parameter is constructed from a temporary [bugprone-return-const-ref-from-parameter,-warnings-as-errors]
  150 |     return document;
      |            ^~~~~~~~

```

bavulapati:
fixes #1657

### DIFF
--- a/src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
+++ b/src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
@@ -74,7 +74,6 @@ using PointerTemplate = GenericPointerTemplate<Pointer>;
 SOURCEMETA_CORE_JSONPOINTER_EXPORT
 auto get(const JSON &document, const Pointer &pointer) -> const JSON &;
 
-// overloaded function
 // Constant reference parameters can accept xvalues which will be destructed
 // after the call. When the function returns such a parameter also as constant
 // reference, then the returned reference can be used after the object it refers
@@ -108,7 +107,6 @@ auto get(JSON &&document, const Pointer &pointer) -> const JSON & = delete;
 SOURCEMETA_CORE_JSONPOINTER_EXPORT
 auto get(const JSON &document, const WeakPointer &pointer) -> const JSON &;
 
-// overloaded function
 // Constant reference parameters can accept xvalues which will be destructed
 // after the call. When the function returns such a parameter also as constant
 // reference, then the returned reference can be used after the object it refers


src/core/jsonpointer/jsonpointer.cc
```diff
@@ -137,15 +137,17 @@ auto try_traverse(const sourcemeta::core::JSON &document,
 
 namespace sourcemeta::core {
 
-auto get(const JSON &document, const Pointer &pointer) -> const JSON & {
+auto get(const JSON &document [[clang::lifetimebound]], const Pointer &pointer)
```
bavulapati, line 140: This seem to be clang specific

src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
```diff
@@ -52,6 +52,12 @@ const WeakPointer empty_weak_pointer;
 /// A JSON Pointer with unresolved wildcards
 using PointerTemplate = GenericPointerTemplate<Pointer>;
 
+// overloaded functions
+auto get(const JSON &&document, const Pointer &pointer)
```
jviotti, line 56: Also, could this be like this:

```suggestion
auto get(JSON &&document, const Pointer &pointer)
```

bavulapati: I don't think so, because the check talks about returning const ref parameter as const ref.
I will try to test it though

src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
```diff
@@ -98,6 +108,16 @@ auto get(const JSON &document, const Pointer &pointer) -> const JSON &;
 SOURCEMETA_CORE_JSONPOINTER_EXPORT
 auto get(const JSON &document, const WeakPointer &pointer) -> const JSON &;
 
+// overloaded function
```
jviotti, line 111:
```suggestion
```

jviotti (approved):

jviotti:
I just removed the `overloaded function` comments (as that's implicit), but otherwise looks good. Thanks!

Issue #1654, opened 2025-05-23
ClangTidy warnings for `bugprone-branch-clone`


jviotti:
```
/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:61:5: error: if with identical then and else branches [bugprone-branch-clone,-warnings-as-errors]
   61 |     if (this->hasher.is_perfect(key_hash)) {
      |     ^
/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:68:7: note: else branch starts here
   68 |     } else {
      |       ^

/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:84:5: error: if with identical then and else branches [bugprone-branch-clone,-warnings-as-errors]
   84 |     if (this->hasher.is_perfect(key_hash)) {
      |     ^
/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:91:7: note: else branch starts here
   91 |     } else {
      |       ^

/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:160:5: error: if with identical then and else branches [bugprone-branch-clone,-warnings-as-errors]
  160 |     if (this->hasher.is_perfect(key_hash)) {
      |     ^
/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:166:7: note: else branch starts here
  166 |     } else {
      |       ^

/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:184:5: error: if with identical then and else branches [bugprone-branch-clone,-warnings-as-errors]
  184 |     if (this->hasher.is_perfect(key_hash)) {
      |     ^
/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:190:7: note: else branch starts here
  190 |     } else {
      |       ^

/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:211:5: error: if with identical then and else branches [bugprone-branch-clone,-warnings-as-errors]
  211 |     if (this->hasher.is_perfect(key_hash)) {
      |     ^
/Users/jviotti/Projects/core/src/core/json/include/sourcemeta/core/json_object.h:217:7: note: else branch starts here
  217 |     } else {
      |       ^

[...]

PR #1734, opened 2025-06-20
style(clang-tidy): Disable bugprone-branch-clone check

bavulapati:
Refs #1654
The check is reporting mostly false positives.
It is creating more noise than value.


jviotti (approved):
Makes sense!

PR #1739, opened 2025-06-20
Attempt to install ClangFormat binaries using CMake

jviotti:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,10 +114,6 @@ jobs:
           HOMEBREW_NO_ANALYTICS: 1
           HOMEBREW_NO_AUTO_UPDATE: 1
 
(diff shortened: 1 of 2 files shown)


bavulapati:
> @bavulapati The problem is that different OSes have different versions of `clang-format` in the package manager, that leads to formatting incompatibilities. That's why we currently pull `clang-format` using `pipx` on CI.
> 
> The way the new function works is that it will try to fetch a pre-built binary, else still fallback to the system one (if any). But for common platforms like macOS, you should always get a pre-built binary hit and then you can forget about version incompatibilities

That makes sense 👍🏾

jviotti:
I think additionally this means I will be able to get formatting work on the MSYS2 build, which I was struggling to make it work before. Maybe as a subsequent PR

PR #1762, opened 2025-06-24
style(clang-tidy): fix bugprone-unchecked-optional-access

bavulapati:


src/core/jsonschema/frame.cc
```diff
@@ -90,9 +91,10 @@ auto find_anchors(const sourcemeta::core::JSON &schema,
       assert(schema.at("id").is_string());
       const sourcemeta::core::URI identifier(schema.at("id").to_string());
       if (identifier.is_fragment_only()) {
-        result.insert(
-            {sourcemeta::core::JSON::String{identifier.fragment().value()},
-             AnchorType::Static});
+        result.insert({sourcemeta::core::JSON::String{
+                           identifier.fragment() ? identifier.fragment().value()
+                                                 : "what should we have here?"},
```
bavulapati, line 96: @jviotti

src/core/jsonschema/frame.cc
```diff
@@ -75,9 +75,10 @@ auto find_anchors(const sourcemeta::core::JSON &schema,
       assert(schema.at("$id").is_string());
       const sourcemeta::core::URI identifier(schema.at("$id").to_string());
       if (identifier.is_fragment_only()) {
-        result.insert(
-            {sourcemeta::core::JSON::String{identifier.fragment().value()},
-             AnchorType::Static});
+        result.insert({sourcemeta::core::JSON::String{
+                           identifier.fragment() ? identifier.fragment().value()
+                                                 : "what should we have here?"},
```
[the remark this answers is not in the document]
jviotti: It is checked, and this other branch is impossible, but I think its just because ClangTidy didn't realise we checked for it using `is_fragment_only` on the `if`

src/core/jsonschema/frame.cc
```diff
@@ -90,9 +91,10 @@ auto find_anchors(const sourcemeta::core::JSON &schema,
       assert(schema.at("id").is_string());
       const sourcemeta::core::URI identifier(schema.at("id").to_string());
       if (identifier.is_fragment_only()) {
```
jviotti, line 93:
```suggestion
      if (identifier.is_fragment_only() && identifier.fragment().has_value()) {
```

Same here. It is a bit redundant, but this could make ClangTidy happy

bavulapati:
@jviotti It would become easier to test these changes after merging #1754

src/core/jsonschema/walker.cc
```diff
@@ -1,3 +1,4 @@
+#include <optional>
```
jviotti, line 1:
```suggestion
```

I don't think you need this include, plus its also in the incorrect place, and without the usual comment

PR #1786, opened 2025-06-26
Install clang-tidy binaries using CMake similar to clang-format

bavulapati:
Refs #1754
This makes it use consistent version of clang-tidy on all the platforms. And also uses fairy latest version of clang-tidy, instead of using old versions shipped along with apple llvm and similar distros.


jviotti (approved):
Great stuff. Thanks a lot.

PR #1792, opened 2025-06-27
Implement a rule to remove empty then or else for the linter

Karan-Palan:
Todo: add more tests for all drafts and edge cases


jviotti:
Awesome! Let me know once its passing and out of draft and I'll take a proper look!

PR #1801, opened 2025-07-02
Clang tidy separate ci job

bavulapati:


bavulapati:
@jviotti it's ready for your review.

bavulapati:
Also supersedes https://github.com/sourcemeta/core/pull/1640

bavulapati:
@jviotti PTAL

.github/workflows/ci.yml
```diff
@@ -161,3 +161,24 @@ jobs:
...
+  clang-tidy:
+    runs-on: ubuntu-latest
+    env:
+      CC: clang
+      CXX: clang++
+    steps:
+      - uses: actions/checkout@v4
+      - run: cmake --version
+      - name: Install llvm
+        run: sudo bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
```
jviotti, line 174: Why do we need this? I thought we tweaked the CMake scripts to conveniently download ClangTidy binaries when configuring?

cmake/common/targets/clang-tidy.cmake
```diff
@@ -1,130 +1,8 @@
...
-
-  # Install the binary
-  file(MAKE_DIRECTORY "${SOURCEMETA_TARGET_CLANG_TIDY_ATTEMPT_INSTALL_OUTPUT_DIRECTORY}")
-  file(COPY "${CLANG_TIDY_BINARY_EXTRACT_DIR}/clang_tidy/data/bin/${CLANG_TIDY_BINARY_NAME}"
-       DESTINATION "${SOURCEMETA_TARGET_CLANG_TIDY_ATTEMPT_INSTALL_OUTPUT_DIRECTORY}")
-  file(CHMOD "${CLANG_TIDY_BINARY_OUTPUT}" PERMISSIONS
-       OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
-  message(STATUS "Installed `clang-tidy` pre-built binary to ${CLANG_TIDY_BINARY_OUTPUT}")
+function(sourcemeta_enable_clang_tidy)
+  find_program(CLANG_TIDY_BIN NAMES clang-tidy)
```
jviotti, line 2: We could still download ClangTidy on demand and set `CMAKE_CXX_CLANG_TIDY` to the downloaded files?

cmake/common/targets/clang-tidy.cmake
```diff
@@ -1,130 +1,8 @@
...
+  set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_BIN};--config-file=${CMAKE_CURRENT_SOURCE_DIR}/.clang-tidy;-header-filter=${CMAKE_CURRENT_SOURCE_DIR}/src/*" PARENT_SCOPE)
```
jviotti, line 3:
```suggestion
  set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_BIN};--config-file=${PROJECT_SOURCE_DIR}/.clang-tidy;-header-filter=${CMAKE_CURRENT_SOURCE_DIR}/src/*" PARENT_SCOPE)
```

I think you want to apply it at the project level?

CMakeLists.txt
```diff
@@ -115,9 +120,10 @@ if(PROJECT_IS_TOP_LEVEL)
     src/*.h src/*.cc
     benchmark/*.h benchmark/*.cc
     test/*.h test/*.cc)
-  sourcemeta_target_clang_tidy(SOURCES src/*.cc)
 endif()
 
+sourcemeta_disable_clang_tidy()
```
jviotti, line 125: Interesting. Disabling looks a bit weird though. Now that I see this I wonder if it is possible to set `CMAKE_CXX_CLANG_TIDY` on a per-target level. If so, we could have a `CLANG_TIDY` option when declaring a library with `sourcemeta_library` (or executable with `sourcemeta_executable`). Then we don't have to worry about unsetting it, as we literally opt into it for specific things we are about?

cmake/common/targets/clang-tidy.cmake
jviotti, line 3: Also, does this work across compilers? i.e. can I enable this if I have ClangTidy available even if using MSVC? Or does it also require the Clang compiler? If so, it might be worth checking for this, so that nobody can accidentally enable it

src/core/json/include/sourcemeta/core/json_array.h
```diff
@@ -56,31 +56,39 @@ template <typename Value> class JSONArray {
   /// Get a mutable end iterator on the array
   auto end() noexcept -> iterator { return this->data.end(); }
   /// Get a constant begin iterator on the array
-  auto begin() const noexcept -> const_iterator { return this->data.begin(); }
+  [[nodiscard]] auto begin() const noexcept -> const_iterator {
```
jviotti, line 59: I'm not sure this is working for me. I removed the `[[nodiscard]]` here, ran `make`, and it all compiles just fine?

cmake/common/targets/clang-tidy.cmake
jviotti: Also keep in mind that this `find_program` invocation succeeds even if `clang-tidy` was not found. Maybe worth putting `REQUIRED`? Otherwise you might end up setting a non-sense `CMAKE_CXX_CLANG_TIDY`?

bavulapati:
Note: We have only enabled single target uuid for clang-tidy checks, for now. We can do it incrementally

cmake/common/targets/clang-tidy.cmake
```diff
@@ -1,130 +1,18 @@
-function(sourcemeta_target_clang_tidy_attempt_install)
```
jviotti, line 1: I think I mentioned this before, but why cannot we pull ClangTidy pre-compiled binaries like before? Otherwise we are back at having it at the system level?

cmake/common/targets/library.cmake
```diff
@@ -115,6 +115,13 @@ function(sourcemeta_library)
     target_include_directories(${TARGET_NAME}
       PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>")
   endif()
+
+  if(SOURCEMETA_LIBRARY_CLANG_TIDY)
+    message(STATUS "setting CXX_CLANG_TIDY target property to ${SOURCEMETA_LIBRARY_CLANG_TIDY} on ${TARGET_NAME}")
+    set_target_properties(${TARGET_NAME}
+      PROPERTIES
+        CXX_CLANG_TIDY "${SOURCEMETA_LIBRARY_CLANG_TIDY}")
```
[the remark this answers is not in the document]
jviotti: Hmmm I find this whole thing still very confusing. Let me try to shoot a WIP PR myself too to play around with this. I still feel there must be a simpler way

jviotti:
Take a look at this: https://github.com/sourcemeta/core/pull/1824. Just a rough experiment but it works on macOS. Seems like we need to tweak the paths a bit more to make it work on Linux, all with the pre-built ClangTidy binaries.

bavulapati:
@jviotti Can you take another look?
It would benefit us to have CI run clang-tidy.

bavulapati:
I've played around with https://cpp-linter.github.io/cpp-linter-action/
CPPLinter also have git pre-commit hooks, which also behaves similar to what we have seen.
CPPLinter also pulls the clang headers but it also misses few headers.

bavulapati:
closing as #1824 is merged.

PR #1803, opened 2025-07-02
style(clang-tidy): fix modernize-use-transparent-functors

bavulapati:


jviotti (approved):
Oh interesting. I didn't know about this one

PR #1810, opened 2025-07-03
style(clang-tidy): fix modernize-use-designated-initializers

bavulapati:


jviotti (approved):
Ah nice! Much cleaner

PR #1811, opened 2025-07-03
style(clang-tidy): fix modernize-use-integer-sign-comparison checks

bavulapati:


src/extension/alterschema/canonicalizer/min_properties_covered_by_required.h
```diff
@@ -1,3 +1,4 @@
+#include <utility>
```
jviotti, line 1: We avoid putting includes in these headers, as they are just composed together in a bigger file. Can you add this here: https://github.com/sourcemeta/core/blob/main/src/extension/alterschema/alterschema.cc#L5-L8 instead?

bavulapati: what do you mean by composed?

bavulapati:
@jviotti PTAL

jviotti (approved):

src/extension/alterschema/canonicalizer/min_properties_covered_by_required.h
jviotti: Just for convenience. There are a ton of rules and a lot of them re-use the exact same includes

PR #1813, opened 2025-07-03
style(clang-tidy): fix modernize-loop-convert fixes

bavulapati:


src/core/uuid/uuid.cc
```diff
@@ -17,8 +17,8 @@ auto uuidv4() -> std::string {
                                                                           15);
   std::string result;
   result.reserve(36);
-  for (std::uint8_t index = 0; index < 16; index++) {
-    if (dash[index]) {
+  for (bool isDash : dash) {
```
jviotti, line 20:
```suggestion
  for (bool is_dash : dash) {
```

Can we use snake_case?

jviotti (approved):

PR #1815, opened 2025-07-03
style(clang-tidy): fix modernize-use-bool-literals

bavulapati:


src/core/uuid/uuid.cc
```diff
@@ -1,3 +1,4 @@
+#include <array>
```
jviotti, line 1: Can you add this include alongside the system includes below and add the usual `std::array` comment?

jviotti (approved):

PR #1819, opened 2025-07-06
Linter: add rule to remove empty then else

Karan-Palan:


src/extension/alterschema/linter/then_else_empty.h
```diff
@@ -0,0 +1,36 @@
+class ThenElseEmpty final : public SchemaTransformRule {
+public:
+  ThenElseEmpty()
+      : SchemaTransformRule{
+            "then_else_empty",
+            "`then` or `else` with an empty schema does not restrict "
+            "validation and is likely ineffective"} {};
```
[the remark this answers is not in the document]
jviotti: Actually, just to make the error simpler by just talking about a single keyword, can you split this rule into two? One of `then` and one for `else` similar to how we catch `then` without `if` and `else` without `if`?

Karan-Palan:
@jviotti 
I have resolved the merge conflicts and addressed the comments. Ptal

src/extension/alterschema/linter/then_empty.h
```diff
@@ -0,0 +1,26 @@
...
+    return contains_any(
+               vocabularies,
+               {"https://json-schema.org/draft/2020-12/vocab/applicator",
+                "https://json-schema.org/draft/2019-09/vocab/applicator",
+                "http://json-schema.org/draft-07/schema#"}) &&
+           schema.is_object() && schema.defines("if") &&
+           schema.defines("then") && is_schema(schema.at("then")) &&
+           ((schema.at("then").is_object() && schema.at("then").empty()) ||
+            (schema.at("then").is_boolean() && schema.at("then").to_boolean() &&
+             !(schema.at("if").is_boolean() && schema.at("if").to_boolean())));
```
jviotti, line 22: Is this line really necessary? I guess the rule applies independently of `if` (and even if `if` is not there?)

src/extension/alterschema/linter/else_empty.h
```diff
@@ -0,0 +1,26 @@
...
+    return contains_any(
+               vocabularies,
+               {"https://json-schema.org/draft/2020-12/vocab/applicator",
+                "https://json-schema.org/draft/2019-09/vocab/applicator",
+                "http://json-schema.org/draft-07/schema#"}) &&
+           schema.is_object() && schema.defines("if") &&
+           schema.defines("else") && is_schema(schema.at("else")) &&
+           ((schema.at("else").is_object() && schema.at("else").empty()) ||
+            (schema.at("else").is_boolean() && schema.at("else").to_boolean() &&
+             !(schema.at("if").is_boolean() && schema.at("if").to_boolean())));
```
jviotti, line 22: Same comment as below

src/extension/alterschema/linter/then_empty.h
jviotti: Ah, is it to avoid conflicts with the other rule that removes dangling `then`/`else` without `if`? I'm still not sure why you check for `if` being a boolean. Maybe just the `schema.defines("if")` is enough?

Karan-Palan: The rule works independently of "if". It was to avoid the conflicts with the other rule. `"schema.defines("if")"` is enough

jviotti (approved):

PR #1824, opened 2025-07-08
Enable ClangTidy alongside compilation in macOS

jviotti:


bavulapati:
@jviotti I updated #1801

PR #1825, opened 2025-07-08
[WIP] feat(alterschema): [linter] create rule to remove empty fragment

Karan-Palan:
Todo: add tests

### DIFF
--- a/src/extension/alterschema/linter/modern_official_dialect_with_empty_fragment.h
+++ b/src/extension/alterschema/linter/modern_official_dialect_with_empty_fragment.h
@@ -1,0 +1,36 @@
+class ModernOfficialDialectWithEmptyFragment final
+    : public SchemaTransformRule {
+public:
+  ModernOfficialDialectWithEmptyFragment()
+      : SchemaTransformRule{
+            "modern_official_dialect_with_empty_fragment",
+            "The official dialect URI of Draft 2019-09 and newer versions must "
+            "not contain the empty fragment"} {};
+
+  [[nodiscard]] auto condition(const sourcemeta::core::JSON &schema,
+                               const sourcemeta::core::JSON &,
+                               const sourcemeta::core::Vocabularies &,
+                               const sourcemeta::core::SchemaFrame &,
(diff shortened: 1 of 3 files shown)


src/extension/alterschema/linter/modern_official_dialect_with_empty_fragment.h
```diff
@@ -0,0 +1,36 @@
+class ModernOfficialDialectWithEmptyFragment final
+    : public SchemaTransformRule {
+public:
+  ModernOfficialDialectWithEmptyFragment()
+      : SchemaTransformRule{
+            "modern_official_dialect_with_empty_fragment",
+            "The official dialect URI of Draft 2019-09 and newer versions must "
```
jviotti, line 7:
```suggestion
            "The official dialect URI of 2019-09 and newer versions must "
```

We don't say draft anymore for the newer ones!

jviotti:
Can you send a separate PR with the official resolver changes? Also we should never be modifying the actual schemas in there, i.e. setting the `$id` accordingly.

src/core/jsonschema/official_resolver.in.cc
```diff
@@ -54,8 +60,10 @@ auto sourcemeta::core::schema_official_resolver(std::string_view identifier)
 
     // Just for compatibility given that this is such a common issue
   } else if (identifier == "https://json-schema.org/draft/2020-12/schema#") {
-    return sourcemeta::core::parse_json(
+    auto schema = sourcemeta::core::parse_json(
         R"EOF(@METASCHEMA_JSONSCHEMA_2020_12@)EOF");
+    schema.at("$id").into(sourcemeta::core::JSON{identifier});
```
jviotti, line 65: We should never mutate the official meta-schemas, as they are supposed to be immutable. What was the intention?

jviotti:
Ah, I see the `assert` you are hitting in `vocabularies()`. I think we should instead change the assertion to parse as URI, canonicalise both, and compare them?

jviotti (approved):
I added a commit to tweak the assertion instead!

PR #1830, opened 2025-07-09
Generalise how to extract any keywords from `allOf`

jviotti:


jviotti:
cc @Karan-Palan FYI: I rephrased one of the rules to extract everything that can be extracted out of any `allOf` for some other work I'm doing right now

PR #1843, opened 2025-07-16
Bring back clang-tidy as a target

bavulapati:


jviotti:
Why do we want to have a target for it though? I thought the whole point was to make it part of the compilation process instead, as it would be able to behave in a better way?

bavulapati:
I was only thinking about the incompatibility with sourcemeta eco system and forgot about the benefits of running it along with compilation. I'll close this tomorrow.

jviotti:
Yeah, sounds good. We can just let it run as part of the compilation process for all the projects!

PR #1844, opened 2025-07-17
Enable clang-tidy at target level

bavulapati:
Follow up https://github.com/sourcemeta/core/pull/1824#issuecomment-3078885364


src/core/yaml/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 sourcemeta_library(NAMESPACE sourcemeta PROJECT core NAME yaml
   PRIVATE_HEADERS error.h
   SOURCES yaml.cc)
+sourcemeta_clang_tidy_attempt_enable(TARGET_NAME "sourcemeta_core_yaml")
```
jviotti, line 4: Can you instead do this by default on the definition of `sourcemeta_library` and `sourcemeta_executable`? Then you don't have to add it for every library

cmake/common/clang-tidy.cmake
```diff
@@ -77,49 +77,53 @@ function(sourcemeta_clang_tidy_attempt_install)
...
       "--extra-arg=${MACOSX_SDK_PATH}"
-      "--extra-arg=-resource-dir=${MACOSX_RESOURCE_PATH}"
-      PARENT_SCOPE)
-  endif()
-endfunction()
+      "--extra-arg=-resource-dir=${MACOSX_RESOURCE_PATH}")
 
-function(sourcemeta_clang_tidy_attempt_disable)
-  unset(CMAKE_CXX_CLANG_TIDY PARENT_SCOPE)
+  set_target_properties("${SOURCEMETA_CLANG_TIDY_TARGET_NAME}"
```
jviotti, line 127: Did you confirm this works? I haven't tried it yet, but when I tried your past PR, it wouldn't seem to like it called ClangTidy at all

bavulapati: This is tested locally

jviotti (approved):
Very nice. I just added a small commit cleaning some stuff up so we can merge it fast.

jviotti:
Ah, I think there is a conflict on the removal of the C++ standard. I'll let you fix that and then we can merge

jviotti (approved):

PR #1847, opened 2025-07-18
Use CMAKE_OSX_SYSROOT

bavulapati:
https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html

This adds the -isysroot flag to the compile commands. This is better because, it adds information to the compilation database. Compilation database is shared across many other tools.


jviotti:
Looks interesting, but what happens if somebody doesn't pass this option? Would ClangTidy be broken? I'll give it a shot

bavulapati:
I would prefer to do it, only if we can make it simple. If we add more conditions, the current one is simpler.
Let's not make it complicated.

PR #1848, opened 2025-07-18
Remove redundant cxx standard

bavulapati:
C++ Standard flag is available in the compilation database.


jviotti (approved):
Cool. Makes sense

PR #1857, opened 2025-07-19
Use json instead of yaml

bavulapati:
Being a repo around JSON, thought of using json here as it is supported. But thinking about the reason for having a yaml instead of JSON, it struck me that json file doesn't allow comments.
What do you think about this limitation?
Does usage of yaml hurt json schema eco system?

### DIFF
--- a/cmake/common/clang-tidy.config
+++ b/cmake/common/clang-tidy.config
@@ -1,19 +1,6 @@
----
-# See https://clang.llvm.org/extra/clang-tidy/index.html
-# First disable all default checks (with -*)
-Checks: '-*,
-          modernize-*'
-# TODO(bavulapati): iterate through the rules and enable them incrementally inorder to send smaller PRs
-          # bugprone-*,-bugprone-branch-clone,-bugprone-easily-swappable-parameters,-bugprone-empty-catch,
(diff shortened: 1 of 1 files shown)


bavulapati:
I want to finally change the file type to .yaml, so that syntax highlighting would be better with LSP.
This change is to just discuss my thoughts and gain more insights.

jviotti (approved):
It is a good question. Lack of comments its the major complain about the JSON grammar, which is why JSON Schema has `$comment`. Other than comments, JSON and YAML is just a matter of taste. You see people passionate about either of them.

I personally prefer JSON as I find that the YAML grammar can get tricky very quickly and it is very easy to mess things up and not really understand why.

In any case,  every JSON document can be transformed to YAML and viceversa so there is no much of a difference (other than comments I guess)

jviotti:
> Does usage of yaml hurt json schema eco system?

On this, I wouldn't say it "hurts", but a common point of confusion is that many beginners don't realise YAML strictly follows the JSON data model and therefore may think that JSON Schema cannot be used to validate YAML. And that instead there exists something called YAML Schema that is a different specification (which is of course false). So there is definitely a bit of confusion out there

PR #1867, opened 2025-07-21
[Linter]: create rule to drop ref siblings in drafts

Karan-Palan:


src/extension/alterschema/linter/draft_ref_siblings.h
```diff
@@ -0,0 +1,52 @@
+class DraftRefSiblings final : public SchemaTransformRule {
+public:
+  DraftRefSiblings()
+      : SchemaTransformRule{
+            "draft_ref_siblings",
+            "The official dialect URI of Draft 7 and older must not contain "
```
jviotti, line 6: What about something like this instead?

> In Draft 7 and older dialects, keywords sibling to `$ref` are never evaluated

src/extension/alterschema/linter/draft_ref_siblings.h
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    if (!schema.is_object() || !schema.defines("$ref")) {
+      return false;
+    }
+
+    for (const auto &entry : schema.as_object()) {
+      const auto &key = entry.first;
+
+      if (key == "$ref" || key == "$schema" || key == "title" ||
```
jviotti, line 36: I'm not sure we should have a whitelist. In theory, every keyword is ignored. We should remove them all

test/alterschema/alterschema_lint_draft7_test.cc
```diff
@@ -1442,3 +1442,84 @@ TEST(AlterSchema_lint_draft7, additional_items_with_schema_items_array_items) {
 
   EXPECT_EQ(document, expected);
 }
+
+TEST(AlterSchema_lint_draft7, draft_ref_siblings_1) {
+  sourcemeta::core::JSON document = sourcemeta::core::parse_json(R"JSON({
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$ref": "#/definitions/foo",
```
jviotti, line 1449: Can you have a test case of `$ref` inside a nested subschema with siblings?

src/extension/alterschema/linter/draft_ref_siblings.h
```diff
@@ -0,0 +1,87 @@
...
+
+    // If at top-level (has $schema), also preserve schema identifiers
+    if (schema.defines("$schema")) {
+      preserve_keywords.insert("$schema");
+
+      // Use walker to check if 'id' is a valid keyword for this dialect
+      if (schema.defines("id")) {
+        const auto id_metadata = walker("id", vocabularies);
+        // If the walker recognizes 'id' as a valid keyword, preserve it
+        if (id_metadata.vocabulary.has_value()) {
```
jviotti, line 43: I guess this works too, but I think its nicer to check the type to not be `Unknown` (https://github.com/sourcemeta/core/blob/0af421fc5b3ca503c8e607a84860b137944e6d52/src/core/jsonschema/include/sourcemeta/core/jsonschema_types.h#L64) here and below

src/extension/alterschema/linter/draft_ref_siblings.h
```diff
@@ -0,0 +1,87 @@
...
+
+    if (!schema.is_object() || !schema.defines("$ref")) {
+      return false;
+    }
+
+    // Determine which keywords to preserve
+    std::set<std::string> preserve_keywords{"$ref"};
+
+    // If at top-level (has $schema), also preserve schema identifiers
+    if (schema.defines("$schema")) {
```
jviotti, line 36: Your approach is not bad, but for a more comprehensive, elegant and robust approach, I propose that you:

- Loop over the object properties in `schema`
- Run the walker and only add to `preserve_keywords` the keywords whose type is `Other`
- Remove the rest

src/extension/alterschema/linter/draft_ref_siblings.h
```diff
@@ -0,0 +1,87 @@
...
+
+  auto transform(JSON &schema) const -> void override {
+    std::set<std::string> preserve_keywords{"$ref"};
+
+    // If at top-level (has $schema), also preserve schema identifiers
+    if (schema.defines("$schema")) {
+      preserve_keywords.insert("$schema");
+
+      // Preserve the appropriate identifier based on what's present
+      if (schema.defines("id")) {
```
jviotti, line 77: Instead of computing everything again, what you can do is turn `preserve_keywords` into a private class member. On the condition handler, you first clear it (in case there were previous runs of the rule), you populate it, and here in the transform handler, you re-use it. I remember I do that same trick in some other rule

src/extension/alterschema/linter/draft_ref_siblings.h
```diff
@@ -0,0 +1,87 @@
...
+                       "http://json-schema.org/draft-00/schema#"})) {
+      return false;
+    }
+
+    if (!schema.is_object() || !schema.defines("$ref")) {
+      return false;
+    }
+
+    // Determine which keywords to preserve
+    std::set<std::string> preserve_keywords{"$ref"};
```
jviotti, line 33: I guess its not much of a difference here, but `std::unordered_set` would likely be faster

src/extension/alterschema/linter/draft_ref_siblings.h
Karan-Palan: Yes that approach is used in `non_applicable_type_specific_keywords`

Karan-Palan:
@jviotti I got this compile error after pulling the recent changes

image

jviotti:
This is a huge one. Thanks a lot!

PR #1868, opened 2025-07-21
Linter:  add rule to remove extra type in propertyNames and remove empty PropertyNames

Karan-Palan:


test/alterschema/alterschema_lint_2019_09_test.cc
```diff
@@ -2032,3 +2032,69 @@ TEST(AlterSchema_lint_2019_09, modern_official_dialect_with_empty_fragment_3) {
...
+      "type": [ "string" ]
+    }
+  })JSON");
+
+  LINT_AND_FIX_FOR_READABILITY(document);
+
+  const sourcemeta::core::JSON expected = sourcemeta::core::parse_json(R"JSON({
+    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "type": "object",
+    "propertyNames": {}
```
jviotti, line 2073: I think we should at the very least take notes to have another rule to remove empty `propertyNames`

PR #1871, opened 2025-07-22
Enable bugprone easily swappable parameters

bavulapati:


bavulapati:
@jviotti resolved conflict

PR #1872, opened 2025-07-22
Enable bugprone-unchecked-optional-access clang-tidy check

bavulapati:
@jviotti We are currently using assert before consuming std::optional's value.
The assert won't be present in production builds.
How should we handle this case? What to do in the case where optional doesn't have value?

### DIFF
--- a/cmake/common/clang-tidy.json
+++ b/cmake/common/clang-tidy.json
@@ -1,7 +1,8 @@
 {
-  "Checks": "-*,
+"Checks": "-*,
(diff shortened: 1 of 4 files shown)


bavulapati:
@jviotti PTAL

jviotti (approved):

jviotti:
Hey @bavulapati , I'm not sure that are you still working on these, so let me close them to reduce noise. Always feel free to re-open if you ever get back to it

PR #1877, opened 2025-07-22
Fix `unnecessary_allof_wrapper_modern` conflict on canonicalise

jviotti:


jviotti:
These are the two times the rule gets called for the first test case:

```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "allOf": [
    {
      "type": "number"
    },
    {
      "type": "string"
    }
  ]
}
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "number",
  "allOf": [
    {
      "type": "string",
      "minLength": 0
    }
  ],
  "multipleOf": 1
}
```

I think the problem is that we record what rule was processed given the pointer of the schema, but when i.e. we remove an entry of an array like `allOf`, we run the same rule against a different schema, but the index is the same, so it fails.

PR #1896, opened 2025-07-27
Enable cppcoreguidelines

bavulapati:


bavulapati:
This builds on top of #1872

PR #1897, opened 2025-07-27
Linter: create rules to add syntaxtical sugar when else is set to false

Karan-Palan:


src/extension/alterschema/linter/then_false.h
jviotti, line 1: While the name of the rules are similar, what they actually do is quite different. Maybe let's do them in 2 PRs?

src/extension/alterschema/linter/else_false.h
```diff
@@ -0,0 +1,36 @@
...
+  condition(const JSON &schema, const JSON &, const Vocabularies &vocabularies,
+            const SchemaFrame &, const SchemaFrame::Location &,
+            const SchemaWalker &, const SchemaResolver &) const
+      -> SchemaTransformRule::Result override {
+    return contains_any(
+               vocabularies,
+               {"https://json-schema.org/draft/2020-12/vocab/applicator",
+                "https://json-schema.org/draft/2019-09/vocab/applicator",
+                "http://json-schema.org/draft-07/schema#"}) &&
+           schema.is_object() && schema.defines("if") &&
```
jviotti, line 17: Should we also check for `then` being missing or being `true`? Otherwise the semantics could break. For example, what if `then` is `false` or `then` is `not: {}` (or any impossible schema)?

src/extension/alterschema/linter/else_false.h
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  auto transform(JSON &schema) const -> void override {
+    const auto if_schema = schema.at("if");
+    schema.erase("if");
+    schema.erase("else");
+    if (if_schema.is_object()) {
+      for (const auto &entry : if_schema.as_object()) {
+        if (!schema.defines(entry.first)) {
+          schema.assign(entry.first, entry.second);
```
jviotti, line 29: There is an `assign_if_missing` method that you could use here to avoid the `defines` check

src/extension/alterschema/linter/else_false.h
```diff
@@ -0,0 +1,36 @@
...
+        }
+      }
+    } else if (if_schema.is_boolean()) {
+      schema = if_schema;
```
jviotti, line 33:
```suggestion
      schema = std::move(if_schema);
```

src/extension/alterschema/linter/else_false.h
```diff
@@ -0,0 +1,36 @@
...
+                "https://json-schema.org/draft/2019-09/vocab/applicator",
+                "http://json-schema.org/draft-07/schema#"}) &&
+           schema.is_object() && schema.defines("if") &&
+           schema.defines("else") && schema.at("else").is_boolean() &&
+           !schema.at("else").to_boolean();
+  }
+
+  auto transform(JSON &schema) const -> void override {
+    const auto if_schema = schema.at("if");
+    schema.erase("if");
```
jviotti, line 24: I'm not sure this is 100% correct. If the parent schema contains certain properties and you thus didn't elevate them, then you cannot remove the `if`? i.e. you have to only remove the `if` if it is empty? 

Can you add tests for this too?

jviotti:

PR #1898, opened 2025-07-27
Linter: create rule to remove properties that are not required

Karan-Palan:


jviotti (changes requested):
I don't think this is a valid one. Let's discuss more in the call. The constraints `properties` add are not related to `required`. If you don't have the properties in `required`, you are still adding valid constraints, just that the properties are optional

PR #1900, opened 2025-07-28
Enable cppcoreguidelines-pro-bounds-pointer-arithmetic

bavulapati:


src/core/yaml/yaml.cc
```diff
@@ -1,4 +1,5 @@
 // See https://pyyaml.org/wiki/LibYAML for basic documentation
+#include <span>
```
jviotti, line 2: Can you put this include after a new line after `<sourcemeta/core/io.h>` and put the usual `// std::span` comment?

jviotti: > Does the llvm clang-tidy check https://clang.llvm.org/extra/clang-tidy/checks/llvm/include-order.html serve this purpose? When I looked at it, it sounded specific to llvm project. Another pair of eyes doesn't harm.

Yeah, that'd be great. I'd be OK with removing all of these comments once we have a check that covers all of these cases and we are confident that works well.

PR #1905, opened 2025-07-30
JSONPointer: add remove method

### LINKED ISSUE #1891: [Feature request] sourcemeta::core::delete(JSON &document, const Pointer &pointer) method to delete object pointed by pointer

dyumin:
Resolves #1891

### DIFF
--- a/src/core/json/include/sourcemeta/core/json_array.h
+++ b/src/core/json/include/sourcemeta/core/json_array.h
@@ -91,6 +91,7 @@ public:
   [[nodiscard]] auto crend() const noexcept -> const_reverse_iterator {
     return this->data.crend();
   }
+
   /// Get array size
   [[nodiscard]] auto size() const noexcept -> size_type {
     return this->data.size();
--- a/src/core/json/include/sourcemeta/core/json_array.h
+++ b/src/core/json/include/sourcemeta/core/json_array.h
@@ -91,6 +91,10 @@ public:
   [[nodiscard]] auto crend() const noexcept -> const_reverse_iterator {
     return this->data.crend();
   }
+  /// Get array size
+  [[nodiscard]] auto size() const noexcept -> size_type {
+    return this->data.size();
+  }
 
 private:
   friend Value;
--- a/src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
+++ b/src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
@@ -313,7 +313,7 @@ auto set(JSON &document, const Pointer &pointer, JSON &&value) -> void;
 /// #include <sstream>
 ///
 /// std::istringstream stream{"[ { \"foo\": 1, \"baz\": 1 }, { \"bar\": 2 } ]"};
-/// sourcemeta::core::JSON document = 
+/// sourcemeta::core::JSON document =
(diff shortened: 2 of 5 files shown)


src/core/jsonpointer/jsonpointer.cc
```diff
@@ -245,6 +245,36 @@ auto set(JSON &document, const Pointer &pointer, JSON &&value) -> void {
   }
 }
 
+template <typename PointerT>
+auto remove_pointer(JSON &document, const PointerT &pointer) -> void {
```
dyumin, line 249: Naming is a bit misleading since it removes pointee not the pointer but I wasn't able to come up with a better one.

src/core/jsonpointer/include/sourcemeta/core/jsonpointer.h
```diff
@@ -303,6 +303,57 @@ auto set(JSON &document, const Pointer &pointer, const JSON &value) -> void;
...
+/// sourcemeta::core::remove(document, pointer);
+/// assert(!document.at(0).defines("foo"));
+/// assert(document.at(0).defines("baz"));
+/// ```
+SOURCEMETA_CORE_JSONPOINTER_EXPORT
+auto remove(JSON &document, const Pointer &pointer) -> bool;
+
+/// @ingroup jsonpointer
+/// Remove a value from a JSON document using a JSON WeakPointer.
+/// Returns true is value is removed, false othervise.
```
jviotti, line 333:
```suggestion
/// Returns true if a value is removed, false otherwise.
```

src/core/json/include/sourcemeta/core/json_array.h
```diff
@@ -91,6 +91,10 @@ template <typename Value> class JSONArray {
   [[nodiscard]] auto crend() const noexcept -> const_reverse_iterator {
     return this->data.crend();
   }
+  /// Get array size
```
jviotti, line 94:
```suggestion

  /// Get array size
```

src/core/jsonpointer/jsonpointer.cc
```diff
@@ -245,6 +245,51 @@ auto set(JSON &document, const Pointer &pointer, JSON &&value) -> void {
...
+    return current.erase(last.to_property()) < current_size;
+  } else {
+    if (current.is_object()) {
+      const auto current_size{current.size()};
+      return current.erase(std::to_string(last.to_index())) < current_size;
+    } else {
+      const auto index{last.to_index()};
+      const auto &array{current.as_array()};
+
+      if (index >= array.size())
```
jviotti, line 274:
```suggestion
      if (index >= array.size()) {
```

Just for consistency

src/core/jsonpointer/jsonpointer.cc
```diff
@@ -245,6 +245,51 @@ auto set(JSON &document, const Pointer &pointer, JSON &&value) -> void {
...
+        return false;
```
jviotti, line 275:
```suggestion
        return false;
      }
```

jviotti:
Looks good. I just left a few minor suggestions.

jviotti:
Actually, let me just add the suggestions myself through the GitHub UI, so we can get it merged and I don't have to bother with a final cycle :)

PR #1906, opened 2025-07-31
Remove non applicable type-specific keywords in terms of `const` and `enum`

jviotti:


jviotti:
@Karan-Palan Take a look at this one, to update the corresponding rule in your docs. This change makes the rule also check for types in `enum` and `const` declarations

PR #1911, opened 2025-08-04
Linter: implement the rule to update definitions to $defs

Karan-Palan:


src/extension/alterschema/linter/definitions_to_defs.h
```diff
@@ -0,0 +1,49 @@
...
+            const sourcemeta::core::Vocabularies &vocabularies,
+            const sourcemeta::core::SchemaFrame &,
+            const sourcemeta::core::SchemaFrame::Location &,
+            const sourcemeta::core::SchemaWalker &,
+            const sourcemeta::core::SchemaResolver &) const
+      -> sourcemeta::core::SchemaTransformRule::Result override {
+    return contains_any(vocabularies,
+                        {"https://json-schema.org/draft/2020-12/vocab/core",
+                         "https://json-schema.org/draft/2019-09/vocab/core"}) &&
+           schema.is_object() && schema.defines("definitions") &&
```
jviotti, line 21: Not for this rule, but another rule that just occurred to me is that you can detect schemas in `definitions` or `$defs` that the rest of the schema is never referencing. It would be an unused schema definition. That also means we can check for empty `definitions` or `$defs` and remove those

test/alterschema/alterschema_lint_2019_09_test.cc
```diff
@@ -2295,6 +2295,68 @@ TEST(AlterSchema_lint_2019_09, property_names_default_1) {
...
+  })JSON");
+
+  LINT_AND_FIX_FOR_READABILITY(document);
+
+  const sourcemeta::core::JSON expected = sourcemeta::core::parse_json(R"JSON({
+    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "type": "object",
+    "properties": {
+      "id": {
+        "$ref": "#/$defs/uuid"
```
jviotti, line 2322: Can you try a more complicated pointer? Like a schema that has a nested schema that defines `$defs` and then it is referenced from the top as `#/foo/bar/$defs/baz` or something?

jviotti: Also with a nested schema resource maybe. Like

```json
{
    "$ref": "nested#/definitions/foo",
    ...
    "$defs": {
      "nested": {
        "$id": "nested",
        ...
        "definitions": {
          "foo": {
            "type": "string"
          }
        }
      }
    }  
}
```

PR #1917, opened 2025-08-06
[WIP] Fix fragment URI escaping

jviotti:


src/core/uri/escaping.cc
```diff
@@ -31,6 +32,19 @@ auto uri_escape(const char character, std::ostream &output,
       return;
     }
   }
+
+  if (mode == URIEscapeMode::Fragment) {
```
jviotti, line 36: TODO: Write test cases for this

PR #1918, opened 2025-08-06
Linter: create rule to to add syntactical sugar when then is set to false

Karan-Palan:
https://github.com/sourcemeta/core/pull/1897#discussion_r2236167053


src/extension/alterschema/linter/then_false.h
```diff
@@ -0,0 +1,29 @@
...
+  condition(const JSON &schema, const JSON &, const Vocabularies &vocabularies,
+            const SchemaFrame &, const SchemaFrame::Location &,
+            const SchemaWalker &, const SchemaResolver &) const
+      -> SchemaTransformRule::Result override {
+    return contains_any(
+               vocabularies,
+               {"https://json-schema.org/draft/2020-12/vocab/applicator",
+                "https://json-schema.org/draft/2019-09/vocab/applicator",
+                "http://json-schema.org/draft-07/schema#"}) &&
+           schema.is_object() && schema.defines("if") &&
```
jviotti, line 18: Same comments as in the other `else_false` PR. I think you will have to check `else` to prevent semantic changes? I think you need to test those too

src/extension/alterschema/linter/then_false.h
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  auto transform(JSON &schema) const -> void override {
+    auto if_schema = schema.at("if");
+    schema.erase("if");
+    schema.erase("then");
+    if (schema.defines("else")) {
+      schema.erase("else");
+    }
+    schema.assign("not", std::move(if_schema));
```
jviotti, line 50: Can you convert the `if` to `not` in place?

src/extension/alterschema/linter/then_false.h
```diff
@@ -0,0 +1,52 @@
...
+
+    return true;
+  }
+
+  auto transform(JSON &schema) const -> void override {
+    auto if_schema = schema.at("if");
+    schema.erase("if");
+    schema.erase("then");
+    if (schema.defines("else")) {
+      schema.erase("else");
```
jviotti, line 48: I think `.erase()` will do nothing if the key doesn't exist, so you can avoid the `defines`

src/extension/alterschema/linter/then_false.h
```diff
@@ -0,0 +1,52 @@
...
+      : SchemaTransformRule{
+            "then_false",
+            "`if: S, then: false` is logically equivalent to `not: S`"} {};
+
+  [[nodiscard]] auto
+  condition(const JSON &schema, const JSON &, const Vocabularies &vocabularies,
+            const SchemaFrame &, const SchemaFrame::Location &,
+            const SchemaWalker &, const SchemaResolver &) const
+      -> SchemaTransformRule::Result override {
+    if (!contains_any(vocabularies,
```
jviotti, line 13: I think this whole condition could be a single chained `&&` line, like some of the other rules. I think it makes it easier to read, otherwise we have to read the inverse conditionals in the whole body. For the `for` loop, you can use `std::all_of`, `std::any_of`, or whatever

src/extension/alterschema/linter/then_false.h
jviotti: I think this comment still applies! You are removing `if` and assigning `not` from scratch. I mean that you can do something like an `.into` on `if` and then `.rename` it to `not`?

src/extension/alterschema/linter/then_false.h
```diff
@@ -0,0 +1,39 @@
+class ThenFalse final : public SchemaTransformRule {
+public:
+  ThenFalse()
+      : SchemaTransformRule{
+            "then_false",
+            "`if: S, then: false` is logically equivalent to `not: S`"} {};
```
jviotti, line 6:
```suggestion
            "`Setting the `then` keyword to the false schema is a convoluted way of negating the `if` subschema, which you can more cleanly represent using the `not` keyword"} {};
```

Just to make it a lot more clear to users

src/extension/alterschema/CMakeLists.txt
```diff
@@ -69,7 +69,8 @@ sourcemeta_library(NAMESPACE sourcemeta PROJECT core NAME alterschema
     linter/definitions_to_defs.h
 
     # Strict
-    strict/required_properties_in_properties.h)
+    strict/required_properties_in_properties.h
+    linter/then_false.h)
```
jviotti, line 73: Same as the other ones. This is not a strict rule

src/extension/alterschema/linter/then_false.h
```diff
@@ -0,0 +1,47 @@
...
+         (schema.at("else").is_boolean() && schema.at("else").to_boolean())) &&
+        (!schema.at("if").is_object() ||
+         std::all_of(schema.at("if").as_object().begin(),
+                     schema.at("if").as_object().end(),
+                     [&schema](const auto &entry) {
+                       return !schema.defines(entry.first);
+                     })));
+
+    if (schema.defines("else")) {
+      return APPLIES_TO_KEYWORDS("if", "then", "else");
```
jviotti, line 34:
```suggestion
      return APPLIES_TO_KEYWORDS("then", "if", "else");
```

Let's keep `then` as the first entry, as that's the main keyword we are discussing here

src/extension/alterschema/linter/then_false.h
jviotti: I think this issue still applies, once more. What you want is to use `.rename`

Karan-Palan:
@jviotti used the new macros and resolved conflicts, ptal

jviotti:

PR #1952, opened 2025-08-21
Don't lint away `allOf` only consisting of 2 or more `$ref`s

jviotti:
This seems to be a recognised pattern out there when extending more than
one schema at the same time.


jviotti:
cc @Karan-Palan For your records. I got a couple of complains about this one.

PR #1958, opened 2025-08-22
Support returning pointers as part of `SchemaTransformRule::Result`

jviotti:


jviotti:
TODO:

- [ ] Make all AlterSchema rules return pointers
- [ ] Test the return pointers of every AlterSchema rule
- [ ] Get rid of `Result::description`
- [ ] Pass `const Result &` to `.transform()` to avoid the `mutable` tricks

PR #1959, opened 2025-08-25
Get rid of `mutable` in AlterSchema rules

jviotti:


jviotti:
cc @Karan-Palan Another one for you to take a look at. Now that conditions can report pointers to the specific locations that the condition matches, we can avoid the `mutable` trick we use in various places by making the transformer take the condition result as a second argument

PR #1973, opened 2025-09-01
Allow `allOf` of multiple branches when the keywords are the same

jviotti:


jviotti:
cc @Karan-Palan See this. I'm extending these rules to not only allow multiple branches containing `$ref` but multiple branches as long as every branch contains exactly the same keywords, as that seems to be a common pattern when defining conjunctions of similar conditions, where there is not "main" one to elevate

PR #1981, opened 2025-09-05
Support custom number of indent spaces in `prettify`

jviotti:


jviotti:
cc @Karan-Palan . I'll add options in the CLI to use this

PR #2023, opened 2025-10-29
[WIP] Improve Unicode handling for regexes

jviotti:


jviotti:
Apparently building Boost::Regex with https://github.com/unicode-org/icu might work

PR #2024, opened 2025-10-29
Revamp the entire regex module based on PCRE2

jviotti:


jviotti:
TODO to self: don't merge this until we confirm it works with the desired Unicode tests?

PR #2026, opened 2025-10-31
Support extracting positions out of YAML parsing

jviotti:


jviotti:
cc @Karan-Palan still needs some work, but almost there

PR #2030, opened 2025-11-05
Implement a `Decimal` numeric module powered by `mpdecimal`

jviotti:


jviotti:
TODO: Have a `src/core/numeric` module that abstracts `mpdecimal` a bit behind an `Decimal` Sourcemeta type or something like that

PR #2040, opened 2025-11-13
perf: Replace unordered_map with bitset for vocabulary lookups

syedazeez337:
- move KnownVocabulary + Vocabularies into their own translation unit
-  add compatibility alias + fixes for callers
- ensure disabled vocabularies still report presence 
- update alter schema helper and walker tests

### DIFF
--- a/src/core/jsonschema/include/sourcemeta/core/jsonschema_vocabularies.h
+++ b/src/core/jsonschema/include/sourcemeta/core/jsonschema_vocabularies.h
@@ -63,7 +63,6 @@ struct SOURCEMETA_CORE_JSONSCHEMA_EXPORT Vocabularies {
   };
 
   // NOTE: Must be kept in sync with the Known enum above
-  // (currently 29 vocabularies)
   static constexpr std::size_t KNOWN_VOCABULARY_COUNT = 29;
 
 public:


src/core/jsonschema/include/sourcemeta/core/jsonschema_vocabularies.h
```diff
@@ -0,0 +1,165 @@
...
+#include <utility>       // std::pair
+#include <vector>        // std::vector
+
+namespace sourcemeta::core {
+
+/// @ingroup jsonschema
+/// Vocabulary enumeration for known JSON Schema vocabularies.
+/// Each vocabulary is represented as a bitflag for efficient storage and
+/// lookup.
+enum class KnownVocabulary : uint32_t {
```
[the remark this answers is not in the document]
jviotti: Also, can you always mention types like `uint32_t` with the `std::` namespace? i.e. `std::uint32_t`? That is more idiomatic

src/core/jsonschema/include/sourcemeta/core/jsonschema_vocabularies.h
```diff
@@ -0,0 +1,165 @@
...
+#pragma warning(push)
+#pragma warning(disable : 4251)
+#endif
+  std::unordered_map<JSON::String, bool> custom;
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif
+
+  /// Default constructor
+  Vocabularies() = default;
```
jviotti: Please add default copy/move constructors/assignment operators too!

src/core/jsonschema/include/sourcemeta/core/jsonschema_vocabularies.h
```diff
@@ -0,0 +1,165 @@
...
+
+  /// Bitflags for disabled known vocabularies (for explicit false values)
+  uint32_t disabled_known = 0;
+
+  /// Fallback storage for custom/unknown vocabularies
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4251)
+#endif
+  std::unordered_map<JSON::String, bool> custom;
```
jviotti: This one mainly. Let's make it private. We don't want consumers to know about this internal detail

src/core/jsonschema/include/sourcemeta/core/jsonschema_vocabularies.h
```diff
@@ -0,0 +1,165 @@
...
+    if (!status.has_value()) {
+      throw std::out_of_range("Vocabulary not found: " + std::string(uri));
+    }
+    return status.value();
+  }
+
+  /// Check if there are no vocabularies
+  /// @return true if no vocabularies are set
+  [[nodiscard]] auto empty() const -> bool {
+    return enabled_known == 0 && disabled_known == 0 && custom.empty();
```
jviotti: Can you check if this can be safely `noexcept`? I think so, but I don't recall if `std::unordered_map::empty` can throw

src/core/jsonschema/include/sourcemeta/core/jsonschema_vocabularies.h
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /// Check if a vocabulary is enabled
+  /// @param uri The vocabulary URI to check
+  /// @return true if the vocabulary is enabled, false otherwise
+  [[nodiscard]] auto contains(std::string_view uri) const -> bool;
+
+  /// Insert a vocabulary with its enabled/disabled status
+  /// @param entry A pair of vocabulary URI and its boolean status
+  auto insert(const std::pair<JSON::String, bool> &entry) -> void;
```
jviotti: Now that we control the class, do we need to make this an `std::pair`? Maybe easier to just take 2 arguments?

src/core/jsonschema/jsonschema.cc
```diff
@@ -438,9 +457,12 @@ auto sourcemeta::core::vocabularies(const SchemaResolver &resolver,
   if (!result.contains(core)) {
```
jviotti: Can we take advantage of your bitsets here? See that we get the core vocabulary for a dialect using `core_vocabulary()`. That one could return a known vocabulary enum member, and thus check containment using the bit maps?

src/core/jsonschema/include/sourcemeta/core/jsonschema_vocabularies.h
```diff
@@ -0,0 +1,165 @@
...
+  Vocabularies(std::initializer_list<std::pair<JSON::String, bool>> init) {
+    for (const auto &entry : init) {
+      insert(entry);
+    }
+  }
+
+  /// Check if a vocabulary is enabled
+  /// @param uri The vocabulary URI to check
+  /// @return true if the vocabulary is enabled, false otherwise
+  [[nodiscard]] auto contains(std::string_view uri) const -> bool;
```
jviotti: Now that you have the bitsets, I think its very important that you have getter methods that can take an enum value? I think that's where the real performance boost will come from: i.e. we can avoid using strings in 90% of the cases

src/core/jsonschema/jsonschema_vocabularies.cc
jviotti: Can we call this `vocabularies.cc`?

src/core/jsonschema/jsonschema_vocabularies.cc
```diff
@@ -0,0 +1,239 @@
...
+#include <vector>   // std::vector
+
+namespace {
+// Map vocabulary URI to its corresponding enum flag
+auto uri_to_known_vocabulary(std::string_view uri)
+    -> std::optional<sourcemeta::core::KnownVocabulary> {
+  using sourcemeta::core::KnownVocabulary;
+
+  // Pre-vocabulary dialects
+  if (uri == "http://json-schema.org/draft-00/schema#")
```
jviotti: Looks good but can you do it the other way around? Let's prefer matching 2020-12 first, then 2019-09, etc. The reason is that latest vocabularies are used more often, so you can short circuit faster. i.e. almost nobody uses Draft 0 right now, but you would still check for that first

src/core/jsonschema/jsonschema_vocabularies.cc
```diff
@@ -0,0 +1,239 @@
...
+}
+
+auto sourcemeta::core::Vocabularies::insert(
+    const std::pair<JSON::String, bool> &entry) -> void {
+  const auto maybe_known = uri_to_known_vocabulary(entry.first);
+  if (maybe_known.has_value()) {
+    const auto flag = static_cast<uint32_t>(maybe_known.value());
+    if (entry.second) {
+      enabled_known |= flag;
+      disabled_known &= ~flag; // Clear from disabled
```
jviotti:
```suggestion
      // Clear from disabled
      disabled_known &= ~flag;
```

Let's put code comments ABOVE the lines for consistency

src/core/jsonschema/jsonschema_vocabularies.cc
```diff
@@ -0,0 +1,239 @@
...
+  }
+  // Fallback to custom vocabularies
+  const auto it = custom.find(std::string{uri});
+  if (it != custom.end()) {
+    return it->second;
+  }
+  return std::nullopt;
+}
+
+auto sourcemeta::core::Vocabularies::merge(const Vocabularies &other) -> void {
```
jviotti: Do we really use this? Can we avoid this method?

jviotti:
I left various comments but the approach is awesome. My big overall comment is that you have a lot of interesting capabilities here, but you are not using them very much yet:

- Internally, you store all official vocabulary URIs using the `enum`
- But then almost every other part of the codebase will still access them using strings

My suggestion: Once you add getters that work on the `enum` values, can you go through the codebase and try to replace any containment check, etc for an official vocabulary URI with the actual enum value? Then we avoid expensive string comparisons. Mainly on `src/extension/alterschema` rules. I think that will speed up the linter by orders of magnitude

jviotti:
Hm, though there are a LOT of places to go through, and many are tricky, like in `frame.cc`. Maybe let's just focus on the foundations merging this in this PR without a lot of extra changes and then we do the other ones in separate PRs? Otherwise this PR will become a monster...

Maybe just add a TODO in the class definition saying that we should start using getters on enum values as much as possible?

src/core/jsonschema/vocabularies.cc
```diff
@@ -0,0 +1,174 @@
...
+  if (uri == "http://json-schema.org/draft-00/schema#")
+    return Vocabularies::Known::JSON_Schema_Draft_0;
+  if (uri == "http://json-schema.org/draft-00/hyper-schema#")
+    return Vocabularies::Known::JSON_Schema_Draft_0_Hyper;
+
+  return std::nullopt;
+}
+} // anonymous namespace
+
+// Vocabularies struct implementations
```
jviotti: Let's remove comment like this that don't add much. i.e. its a bit of an obvious statement. In general, assuming you used Claude Code, I saw it do this a lot: just drop obvious comments everywhere. You might need to tell it not to 😅

src/core/jsonschema/vocabularies.cc
```diff
@@ -0,0 +1,174 @@
...
+  }
+}
+
+auto sourcemeta::core::Vocabularies::contains(
+    std::string_view uri) const noexcept -> bool {
+  const auto maybe_known = uri_to_known_vocabulary(uri);
+  if (maybe_known.has_value()) {
+    return this->contains(maybe_known.value());
+  }
+  // Fallback to custom vocabularies - check if present
```
jviotti: I.e. this is another obvious comment

PR #2145, opened 2025-12-26
Fix incorrect instance locations on `$ref` with `allOf`

jviotti:


jviotti:
TODO: Seems to loop indefinitely in some schemas?

Issue #1975, opened 2025-09-02
Linter: Future Enhancements and follow up tasks


Karan-Palan:
This issue tracks leftover tasks or future enhancements for the JSON Schema linter and canonicalizer. These were identified during GSoC work and should be considered for future development.

### 1. Documentation

* [ ] Move existing rule documentations to the **JSON Schema CLI** repo for now, so they are more visible and semi-official until we upstream them properly.

### 2. Reference Handling

* [ ] Break schemas when necessary and use `rereference()` in rules that transform applicators, to ensure `$ref` pointers stay correct.

### 3. Non-auto-fixable Rules

Implement rules that cannot be safely auto-fixed, but should instead produce a clear user-friendly lint message.

- [ ]  `conflicting_readonly_writeonly.json`
- [ ] `const_and_enum_conflict.json`
- [x]  `examples_must_validate.json`
- [ ] `forbid_empty_enum.json`
- [ ]  `format_type_mismatch.json`
- [ ]  `incoherent_exclusive_limits.json`
- [ ] `incoherent_inclusive_limits.json`
- [ ]  `mincontains_greater_than_maxcontains.json`
- [ ]  `mincontains_zero_without_maxcontains.json`
- [ ]  `no_actionable_text_in_title_description.json`
- [ ]  `oneof_min_branches.json`
- [ ]  `require_absolute_draft_uri.json`
- [x]  `require_description_title.json`
- [ ] `require_schema_declaration.json`
- [ ]  `top_level_id_must_be_absolute.json`
- [ ]  `unbounded_collection_limits.json`
- [ ]  `unique_branches_in_oneof.json`
- [x]  `unresolved_ref_pointer.json`
- [x] `$schema` sibling to `$ref` in JSON Schema Draft 7 and older (we can't just remove it otherwise the linter evaluation breaks given it doesn't know the dialect?)
- [ ] Only use `$anchor` in a saved definition (not randomly)
- [ ] `$ref` must not point to a random schema location, but to `$defs`, or an anchor, or a schema resource

### 5. Autofixable rules

- [x] Remove unused non-referenced definitions
- [x] Flag references to non-existent locations (remove the `$ref`)
- [x] Keywords in `allOf` branches that already exist outside are duplicating work and can be removed
- [ ] `properties` in an `allOf` branch while the outer schema has `additionalProperties: false` and no `properties`/`patternProperties` can be removed. This is one case where we know keywords would never match. Like dead code?
[...]

PR #2338, opened 2026-04-08
feat(linter): add format_type_mismatch rule to detect invalid format usage

Vaibhav701161:
## Summary

This PR introduces a new lint rule: `format_type_mismatch`.

The rule detects schemas where the `format` keyword is used alongside a non-string `type`.

```json
{
  "type": "integer",
  "format": "email"
}
```

In JSON Schema, `format` validation is defined for string instances. Using it with other types does not have any effect and is typically an authoring mistake.

This rule highlights such cases to improve schema correctness and clarity.

The rule emits a diagnostic when:

* `format` exists
* `type` exists
* `type` is a single string value
* `type` is not `"string"`

The diagnostic points to the `format` keyword location.

This rule is non auto-fixable because modifying either `type` or `format` would require assumptions about the author’s intent.

---

## Implementation

The rule follows the existing alterschema linter architecture:

* implemented as a header-only rule in
  `src/extension/alterschema/linter/format_type_mismatch.h`
* registered in `alterschema.cc`
* added to the `SOURCES` list in the alterschema CMake configuration

The rule checks:

* schema node is an object
* `format` is defined and is a string
* `type` is defined and is a string
* `type != "string"`

If these conditions are met the rule returns:

```
APPLIES_TO_KEYWORDS("format")
```

---

## Design Notes

This rule complements the existing `non_applicable_type_specific_keywords` rule.

While the existing rule detects keywords that are not applicable to a given type, this rule provides a more explicit and focused diagnostic specifically for misuse of the `format` keyword.

This improves clarity for users by directly pointing out incorrect `format` usage rather than reporting it as a generic type incompatibility.

The rule intentionally skips cases where `type` is an array (e.g., `["string", "integer"]`), since `format` may still apply when the instance is a string.

The rule applies across all supported drafts where both `type` and `format` are valid keywords.

---

## Tests

Tests were added across all supported dialects using the existing lint testing utilities:

* Draft 3, Draft 4, Draft 6, Draft 7
* Draft 2019-09
* Draft 2020-12

The following scenarios are covered:

* Rule fires when `format` is used with non-string types
* Rule does not fire when `type` is `"string"`
* Rule does not fire when `format` is missing
* Rule does not fire when `type` is missing
* Rule does not fire when `type` is an array including `"string"`
* Rule fires inside nested subschemas (`properties`, `items`, etc.)
* Rule correctly handles `$ref`:

  * pointing to affected subschemas
  * pointing inside affected subschemas

Tests are structured consistently with existing linter rule tests and account for interactions with other rules where applicable.

---

## Related Work

Refs: #1975

### DIFF
--- a/src/extension/alterschema/CMakeLists.txt
+++ b/src/extension/alterschema/CMakeLists.txt
@@ -85,6 +85,7 @@ sourcemeta_library(NAMESPACE sourcemeta PROJECT core NAME alterschema
     linter/enum_to_const.h
     linter/equal_numeric_bounds_to_const.h
     linter/forbid_empty_enum.h
+    linter/format_type_mismatch.h
     linter/incoherent_min_max_contains.h
     linter/invalid_external_ref.h
     linter/items_array_default.h
--- a/src/extension/alterschema/alterschema.cc
+++ b/src/extension/alterschema/alterschema.cc
@@ -114,6 +114,7 @@ inline auto APPLIES_TO_POINTERS(std::vector<Pointer> &&keywords)
 #include "linter/enum_to_const.h"
 #include "linter/equal_numeric_bounds_to_const.h"
 #include "linter/forbid_empty_enum.h"
+#include "linter/format_type_mismatch.h"
 #include "linter/incoherent_min_max_contains.h"
 #include "linter/invalid_external_ref.h"
 #include "linter/items_array_default.h"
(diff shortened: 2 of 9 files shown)


jviotti:
Nice one! Though note that we recently move the `alterschema` module down to Blaze: https://github.com/sourcemeta/blaze/tree/main/src/alterschema. Do you mind sending it over there instead?

jviotti:

PR #2349, opened 2026-04-17
feat(frame): Add PositionTracker for Uri Collision Errors

MohamedKamal000:
this helps to solve https://github.com/sourcemeta/jsonschema/issues/321
i have used the `PositionTracker` to get the line and column, Im not sure if there is another way to do the same thing
i think (if thats the right direction to solve this) that we might need to refactor the `to_json` so it uses the internal `PositionTracker` instead  
i will start writing some unit tests just want to make sure that this is the right direction

### DIFF
--- a/src/core/jsonschema/frame.cc
+++ b/src/core/jsonschema/frame.cc
@@ -266,18 +266,27 @@ auto throw_already_exists(const sourcemeta::core::JSON::String &uri) -> void {
                                            "Schema identifier already exists");
 }
 
-auto store(sourcemeta::core::SchemaFrame::Locations &frame,
-           const sourcemeta::core::SchemaReferenceType type,
-           const sourcemeta::core::SchemaFrame::LocationType entry_type,
-           sourcemeta::core::JSON::String uri, const std::string_view base,
-           const sourcemeta::core::WeakPointer &pointer_from_root,
-           const std::size_t relative_pointer_offset,
-           const std::string_view dialect,
-           const sourcemeta::core::SchemaBaseDialect base_dialect,
-           const std::optional<sourcemeta::core::WeakPointer> &parent,
-           const bool property_name, const bool orphan,
-           const bool ignore_if_present = false,
-           const bool already_canonical = false) -> void {
+[[noreturn]]
+auto throw_already_exists(const sourcemeta::core::JSON::String &uri,
+                          const std::uint64_t line, const std::uint64_t column)
+    -> void {
+  throw sourcemeta::core::SchemaUriCollisionError(
+      uri, line, column, "Schema identifier already exists");
+}
+
+auto store(
+    sourcemeta::core::SchemaFrame::Locations &frame,
+    const sourcemeta::core::SchemaReferenceType type,
+    const sourcemeta::core::SchemaFrame::LocationType entry_type,
+    sourcemeta::core::JSON::String uri, const std::string_view base,
+    const sourcemeta::core::WeakPointer &pointer_from_root,
+    const std::size_t relative_pointer_offset, const std::string_view dialect,
+    const sourcemeta::core::SchemaBaseDialect base_dialect,
+    const std::optional<sourcemeta::core::WeakPointer> &parent,
+    const bool property_name, const bool orphan,
+    const std::optional<sourcemeta::core::PointerPositionTracker> &tracker,
+    const bool ignore_if_present = false, const bool already_canonical = false)
+    -> void {
   auto canonical{already_canonical ? std::move(uri)
                                    : sourcemeta::core::URI::canonicalize(uri)};
   auto [iterator, inserted] =
@@ -292,6 +301,14 @@ auto store(sourcemeta::core::SchemaFrame::Locations &frame,
                      .property_name = property_name,
                      .orphan = orphan}});
   if (!ignore_if_present && !inserted) {
+    if (tracker) {
+      const auto position =
+          tracker->get(sourcemeta::core::to_pointer(pointer_from_root));
+      throw_already_exists(iterator->first.second,
+                           std::get<0>(position.value()),
+                           std::get<1>(position.value()));
+    }
+
     throw_already_exists(iterator->first.second);
   }
 
@@ -430,6 +447,10 @@ auto SchemaFrame::to_json(
   return root;
 }
 
+auto SchemaFrame::add_tracker(const PointerPositionTracker &tracker) -> void {
+  this->position_tracker_ = tracker;
+}
+
 auto SchemaFrame::analyse(const JSON &root, const SchemaWalker &walker,
                           const SchemaResolver &resolver,
                           std::string_view default_dialect,
@@ -499,7 +520,8 @@ auto SchemaFrame::analyse(const JSON &root, const SchemaWalker &walker,
       store(this->locations_, SchemaReferenceType::Static,
             SchemaFrame::LocationType::Resource, default_id_canonical,
             this->root_, path, path.size(), root_dialect,
-            root_base_dialect.value(), std::nullopt, false, false);
+            root_base_dialect.value(), std::nullopt, false, false,
+            this->position_tracker_);
 
       base_uris.insert({path, {root_id.value(), default_id_canonical}});
     }
@@ -606,6 +628,12 @@ auto SchemaFrame::analyse(const JSON &root, const SchemaWalker &walker,
                 this->locations_.find({SchemaReferenceType::Static, new_id})};
             if (maybe_match != this->locations_.cend() &&
                 maybe_match->second.pointer != common_pointer_weak) {
+              if (this->position_tracker_) {
+                const auto position =
+                    position_tracker_->get(to_pointer(common_pointer_weak));
+                throw_already_exists(new_id, std::get<0>(position.value()),
+                                     std::get<1>(position.value()));
+              }
               throw_already_exists(new_id);
             }
 
@@ -618,7 +646,7 @@ auto SchemaFrame::analyse(const JSON &root, const SchemaWalker &walker,
                     common_pointer_weak, common_pointer_weak.size(),
                     entry.common.dialect, entry.common.base_dialect.value(),
                     common_parent, entry.common.property_name,
-                    entry.common.orphan);
+                    entry.common.orphan, this->position_tracker_);
             }
 
             auto base_uri_match{base_uris.find(common_pointer_weak)};
@@ -687,7 +715,7 @@ auto SchemaFrame::analyse(const JSON &root, const SchemaWalker &walker,
                   common_pointer_weak, bases.second.size(),
                   entry.common.dialect, entry.common.base_dialect.value(),
                   common_parent, entry.common.property_name,
-                  entry.common.orphan);
+                  entry.common.orphan, this->position_tracker_);
           }
 
           if (type == AnchorType::Dynamic || type == AnchorType::All) {
@@ -696,7 +724,7 @@ auto SchemaFrame::analyse(const JSON &root, const SchemaWalker &walker,
                   common_pointer_weak, bases.second.size(),
                   entry.common.dialect, entry.common.base_dialect.value(),
                   common_parent, entry.common.property_name,
-                  entry.common.orphan);
+                  entry.common.orphan, this->position_tracker_);
 
             // Register a dynamic anchor as a static anchor if possible too
             if (entry.common.vocabularies.contains(
@@ -706,7 +734,7 @@ auto SchemaFrame::analyse(const JSON &root, const SchemaWalker &walker,
                     common_pointer_weak, bases.second.size(),
                     entry.common.dialect, entry.common.base_dialect.value(),
                     common_parent, entry.common.property_name,
-                    entry.common.orphan, true);
+                    entry.common.orphan, this->position_tracker_, true);
             }
           }
         } else {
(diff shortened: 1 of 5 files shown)


src/core/jsonschema/include/sourcemeta/core/jsonschema_frame.h
```diff
@@ -154,6 +154,8 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrame {
                std::string_view default_id = "",
                const Paths &paths = {empty_weak_pointer}) -> void;
 
+  auto add_tracker(const PointerPositionTracker &tracker) -> void;
```
jviotti, line 157: Let's not couple the `SchemaFrame` with the tracker. The way we do this in other cases is that the exception does not add line/column information at all. It just provides the pointer. Then applications that care about the line numbers can catch the error, and with the pointer + tracker, derive the line numbers

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,40 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
   const char *message_;
 };
 
+/// @ingroup jsonschema
+/// An error that represents a schema Uri Collision error
+class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaUriCollisionError
```
jviotti, line 242: Can we make this specifically focused on anchor collisions? i.e. `SchemaAnchorCollisionError`? So it fires for _just_ that type of collision?

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,40 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
...
+    : public std::exception {
+public:
+  SchemaUriCollisionError(const std::string_view identifier,
```
jviotti, line 245: As per my comment below, please take and expose the `Pointer`, rather than line/column

jviotti: I also think it would be a nice bonus if we can have i.e. an `.other()` method that gives you the `Pointer` of the _other_ anchor location causing the conflict?

src/core/jsonschema/frame.cc
```diff
@@ -266,18 +266,27 @@ auto throw_already_exists(const sourcemeta::core::JSON::String &uri) -> void {
...
-           const sourcemeta::core::SchemaBaseDialect base_dialect,
-           const std::optional<sourcemeta::core::WeakPointer> &parent,
-           const bool property_name, const bool orphan,
-           const bool ignore_if_present = false,
-           const bool already_canonical = false) -> void {
+[[noreturn]]
+auto throw_already_exists(const sourcemeta::core::JSON::String &uri,
+                          const std::uint64_t line, const std::uint64_t column)
+    -> void {
+  throw sourcemeta::core::SchemaUriCollisionError(
```
jviotti, line 273: Let's make this focused on anchors instead?

src/core/jsonschema/include/sourcemeta/core/jsonschema_frame.h
MohamedKamal000: ooh ok it makes sense now

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
MohamedKamal000: ok then, will make it only for AnchorCollisions

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
MohamedKamal000: so if i understand correctly, we would have two pointers one where it was already exist and the other which causes the collision ?
and the message error would look like 

Schema identifier error ...
Anchor <anchor name> already exist at line, column
Redefined at line, column 

something around this

src/core/jsonschema/frame.cc
MohamedKamal000: ok
should i also leave the override of the function and just make throw `SchemaAnchorCollisionError`
or maybe change its name to something like throw_anchor_collision ?

MohamedKamal000:
@jviotti  what do you think about it now ?
also there was some tests related to anchors collision already i edited them so they expect to throw the new Exception 
do you think we need tests for this ? 
thanks for your reviews btw

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,41 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
   const char *message_;
 };
 
+/// @ingroup jsonschema
+/// An error that represents a schema Uri Collision error
```
jviotti:
```suggestion
/// An error that represents a schema anchor collision error
```

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,41 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
...
+class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaAnchorCollisionError
+    : public std::exception {
+public:
+  SchemaAnchorCollisionError(const std::string_view identifier,
+                             const char *message, Pointer collision_location,
```
jviotti:
```suggestion
                             const char *message, Pointer location,
```

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,41 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
...
+                             Pointer other)
+      : identifier_{identifier}, message_{message},
+        collision_location_(std::move(collision_location)),
```
jviotti:
```suggestion
        location_(std::move(location)),
```

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,41 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
...
+  [[nodiscard]] auto what() const noexcept -> const char * override {
+    return this->message_;
+  }
+
+  [[nodiscard]] auto identifier() const noexcept -> std::string_view {
+    return this->identifier_;
+  }
+
+  [[nodiscard]] auto location() const noexcept -> const Pointer & {
+    return collision_location_;
```
jviotti:
```suggestion
    return this->location_;
```

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,41 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
...
+  }
+
+  [[nodiscard]] auto other() const noexcept -> const Pointer & {
+    return other_;
```
jviotti:
```suggestion
    return this->other_;
```

Remember to be consistent!

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,41 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
...
+  }
+
+private:
+  std::string identifier_;
+  const char *message_;
+  Pointer collision_location_;
```
jviotti:
```suggestion
  Pointer location_;
```

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,41 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
...
+  Pointer other_; // where the first anchor appeared
```
jviotti:
```suggestion
  Pointer other_;
```

This comment is not really adding anything not obvious

src/core/jsonschema/frame.cc
```diff
@@ -266,6 +266,17 @@ auto throw_already_exists(const sourcemeta::core::JSON::String &uri) -> void {
                                            "Schema identifier already exists");
 }
 
+[[noreturn]]
+auto throw_already_exists_anchor(
```
jviotti: This is a one-line function only used in a single place. Can we just inline it whenever appropriate?

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,41 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
...
+public:
+  SchemaAnchorCollisionError(const std::string_view identifier,
+                             const char *message, Pointer collision_location,
+                             Pointer other)
+      : identifier_{identifier}, message_{message},
+        collision_location_(std::move(collision_location)),
+        other_(std::move(other)) {}
+
+  [[nodiscard]] auto what() const noexcept -> const char * override {
+    return this->message_;
```
jviotti: I think we can hardcode the message here. It is only one message anyway

test/jsonschema/jsonschema_frame_2019_09_test.cc
```diff
@@ -815,7 +815,7 @@ TEST(JSONSchema_frame_2019_09, static_anchor_override) {
       sourcemeta::core::SchemaFrame::Mode::References};
   EXPECT_THROW(frame.analyse(document, sourcemeta::core::schema_walker,
                              sourcemeta::core::schema_resolver),
-               sourcemeta::core::SchemaFrameError);
+               sourcemeta::core::SchemaAnchorCollisionError);
```
jviotti: Can you expand these tests to not do `EXPECT_THROW` but rather a full `try` / `catch` so we can assert on the properties of the error? I believe we do this in other frame tests. Look for try/catch to see the pattern

test/jsonschema/jsonschema_frame_2019_09_test.cc
```diff
@@ -2233,7 +2233,7 @@ TEST(JSONSchema_frame_2019_09, recursive_anchor_conflict) {
...
 }
 
 TEST(JSONSchema_frame_2019_09, invalid_recursive_ref) {
```
jviotti, line 2255: Same as above

test/jsonschema/jsonschema_frame_2020_12_test.cc
```diff
@@ -851,7 +851,7 @@ TEST(JSONSchema_frame_2020_12, static_anchor_override) {
       sourcemeta::core::SchemaFrame::Mode::References};
   EXPECT_THROW(frame.analyse(document, sourcemeta::core::schema_walker,
                              sourcemeta::core::schema_resolver),
-               sourcemeta::core::SchemaFrameError);
+               sourcemeta::core::SchemaAnchorCollisionError);
```
jviotti: Same as above. Let's do the full `try`/`catch` pattern in these 3 to assert on the error data

jviotti (changes requested):
Very nice. Much better. Some minor comments remaining!

MohamedKamal000:
@jviotti done

src/core/jsonschema/include/sourcemeta/core/jsonschema_error.h
```diff
@@ -237,6 +237,38 @@ class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaFrameError
...
+class SOURCEMETA_CORE_JSONSCHEMA_EXPORT SchemaAnchorCollisionError
+    : public std::exception {
+public:
+  SchemaAnchorCollisionError(const std::string_view identifier,
+                             Pointer location, Pointer other)
+      : identifier_{identifier}, location_(std::move(location)),
+        other_(std::move(other)) {}
+
+  [[nodiscard]] auto what() const noexcept -> const char * override {
+    return "Schema identifier already exists";
```
jviotti:
```suggestion
    return "Schema anchor already exists";
```

src/core/jsonschema/frame.cc
```diff
@@ -266,6 +266,16 @@ auto throw_already_exists(const sourcemeta::core::JSON::String &uri) -> void {
                                            "Schema identifier already exists");
 }
 
+[[noreturn]]
+inline auto throw_already_exists_anchor(
```
jviotti: This is still pending. Why not inline in the calling place?

MohamedKamal000: sorry but i thought what you meant is add the inline keyword, I'm still not sure what you mean by inline in the calling place 
do you mean just remove the function and use its body instead where i called it, since it is already used in one place of the code only ?

jviotti: Yes. Because it is just a `throw`, you can remove the helper function and just do the `throw` on that place. Simpler

MohamedKamal000: ok then

jviotti (approved):
Amazing! Thanks!

MohamedKamal000:
should i do a follow up pr in the `jsonschema` repo to make use of this new exception now ? @jviotti

jviotti:
Ah, I already had done so right after merging your PR: https://github.com/sourcemeta/jsonschema/pull/735

MohamedKamal000:
oh ok cool, will take a look at it to better understand how we can make use of the pointers. Thanks for your reviews !

PR #2354, opened 2026-04-24
Fix URI parser: IPv6 structural validation and pct-encoded case sensitivity

### LINKED ISSUE #2353: URI parser: 7 mismatches found against RFC 3986 test suite
## Bug Report

**Component:** URI Parser  
**Found via:** [JSON Schema Test Suite PR #895](https://github.com/json-schema-org/JSON-Schema-Test-Suite/pull/895)  
**Test file:** draft2020-12 URI suite (240 tests, 7 mismatches)

---

### False Positives - parser accepts invalid URIs (6)

- [ ] **`http://[]` - empty brackets accepted**  
  RFC 3986 §3.2.2: `IP-literal = "[" ( IPv6address / IPvFuture ) "]"`; empty content satisfies neither alternative.

- [ ] **`http://[2001:db8::00000]` - five-digit h16 group accepted**  
  RFC 3986 §3.2.2: `h16 = 1*4HEXDIG`; five hex digits in one group exceed the maximum of four.

- [ ] **`http://[2001::db8::1]` - double `::` accepted**  
  RFC 3986 §3.2.2: each `IPv6address` alternative allows at most one `::`; two occurrences match no alternative.

- [ ] **`http://[1:2:3:4:5:6:7]` - seven groups without `::` accepted**  
  RFC 3986 §3.2.2: without `::`, the full `IPv6address` form requires exactly 8 groups; 7 groups are invalid.

- [ ] **`http://[1:2:3:4:5:6:7:8:9]` - nine groups accepted**  
  RFC 3986 §3.2.2: no `IPv6address` alternative allows more than 8 groups.

- [ ] **`http://[::ffff:1.2.3.256]` - IPv4 octet 256 out of range accepted**  
  RFC 3986 §3.2.2: `IPv6address` `ls32` reuses `IPv4address`; `dec-octet` alt5 = `"25" %x30-35` has a maximum of 255; 256 matches no alternative.

---

### False Negatives - parser rejects valid URIs (1)

- [ ] **`http://a.com/%aF` - mixed-case hex in percent-encoding rejected**  
  RFC 3986 §2.1: `pct-encoded = "%" HEXDIG HEXDIG`; HEXDIG is defined in RFC 5234 core rules as case-insensitive, so lowercase hex digits are valid.

---

### Notes

All 7 cases have a `comment` field in the test suite with the exact ABNF citation. The 233 remaining tests pass correctly.

AcEKaycgR:
Fixes #2353

### IPv6 false positives (6 cases)
`parse_ipv6()` previously accepted any sequence of hex digits, colons, and dots inside brackets without validating RFC 3986 §3.2.2 / RFC 4291 structural rules. After the character-level scan, the extracted address is now passed to `is_ipv6()` which already enforces all the rules correctly:

- `http://[]` - empty brackets
- `http://[2001:db8::00000]` - h16 group exceeds 4 hex digits
- `http://[2001::db8::1]` - more than one `::`
- `http://[1:2:3:4:5:6:7]` - 7 groups without `::`
- `http://[1:2:3:4:5:6:7:8:9]` - 9 groups
- `http://[::ffff:1.2.3.256]` - embedded IPv4 octet out of range

### Percent-encoding false negative (1 case)
`validate_percent_encoded_utf8()` was treating bytes like `0xAF` as bare UTF-8 continuation bytes and throwing. RFC 3986 §2.1 only requires two valid `HEXDIG` after `%`; UTF-8 validity of the encoded octet is not required. The function now returns after confirming two valid hex digits.

- `http://a.com/%aF` - mixed-case hex rejected incorrectly

## Tests
7 regression tests added to `uri_parse_test.cc`, one per fixed case.
All 628 existing URI unit tests continue to pass.

Its ready for review @jviotti

### DIFF
--- a/src/core/uri/parse.cc
+++ b/src/core/uri/parse.cc
@@ -1,3 +1,4 @@
+#include <sourcemeta/core/ip.h>
 #include <sourcemeta/core/uri.h>
 
 #include "escaping.h"
@@ -22,9 +23,7 @@ using namespace sourcemeta::core;
 auto validate_percent_encoded_utf8(const std::string_view input,
                                    std::string_view::size_type position)
     -> std::string_view::size_type {
-  if (input[position] != URI_PERCENT) {
-    return 3;
-  }
+  assert(input[position] == URI_PERCENT);
 
   if (position + 2 >= input.size()) [[unlikely]] {
     throw sourcemeta::core::URIParseError{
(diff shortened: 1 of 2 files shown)


src/core/uri/parse.cc
```diff
@@ -238,10 +182,16 @@ auto parse_ipv6(const std::string_view input,
         static_cast<std::uint64_t>(start + 1)};
   }
 
+  const auto literal{input.substr(start + 1, position - start - 1)};
+  if (!is_ipvfuture && !sourcemeta::core::is_ipv6(literal)) [[unlikely]] {
```
jviotti, line 186: Ahh good catch. We developed the `ip` module very recently. Makes sense using here now! 💯

jviotti (approved):
Impeccable PR. Great job. Thanks!

PR #2465, opened 2026-05-30
json_hash: SIMD-accelerate PropertyHashJSON::perfect

SouptikH:
## Summary

Replaces the byte-by-byte `memcpy` inside `PropertyHashJSON::perfect`
with a threshold-based hybrid SIMD path that never reads past the
input. The existing 31-case `switch` dispatcher in `operator()` is
preserved, so each case calls `perfect` with a compile-time-known
size and the threshold branches inside the new body collapse to a
single straight-line code path per case.

| Input size | Path | Why |
| --- | --- | --- |
| 1 .. 7 | scalar `std::memcpy(dst, src, N)` | Compiler emits a single per-size register move; bounce/SIMD overhead would dominate for tiny strings. |
| 8 .. 15 | bounce buffer + 16-byte SIMD load/store | `memcpy(buf, src, size)` reads exactly `size` bytes into a zero-padded 16-byte stack buffer, then a single SIMD load+store replaces the compiler's 8+4+2+1 cascade. |
| 16 .. 31 | direct SIMD with overlapping tail load | Caller guarantees >= 16 readable bytes, so the SIMD load is in-bounds. Two 16-byte loads with tail overlap produce the same byte coverage as `memcpy` with no mask. |
| >= 32 | unchanged collision-tag fallback | Original path. |

Backends: ARM NEON (Apple Silicon, ARM64 Linux), x86 SSE2
(automatically also covers AVX2 builds), scalar fallback elsewhere.

## Why this is safe

- The first byte of `result` (low byte of `hash.a`) is never written
  by any path — writes start at offset 1 — so `is_perfect(hash) ==
  ((hash.a & 255) == 0)` keeps its semantics.
- For sizes 8 .. 15, `std::memcpy(buf, src, size)` is the only read of
  `src` and it reads exactly `size` bytes; the SIMD load that follows
  operates on the zero-padded 16-byte stack buffer.
- For sizes 16 .. 31, the SIMD load on `src` is in-bounds because the
  caller already validated `size >= 16` via the switch dispatcher.
  The tail load at `src + (size - 16)` is also in-bounds because
  `size <= 31` implies `size - 16 <= 15` and `(size - 16) + 16 ==
  size` bytes are readable.
- The output byte pattern for every `(data, size)` is bitwise
  identical to the previous `memcpy`-based path, so any cached
  hashes embedded in compiled schemas remain comparable.

## ASAN

Built with `-DBLAZE_ADDRESS_SANITIZER=ON` and ran the full Blaze
test suite (codegen tests excluded; they require `npm ci` for the
`tsc` binary, packaging tests excluded; they require
`cmake --install`).

```
100% tests passed, 0 tests failed out of 19
Total Test time (real) =  17.18 sec
```

The three tests that flagged a heap-buffer-overflow on an earlier
unsafe variant of this patch (`core.json`, `core.jsonpointer`,
`core.uritemplate`) are all green under ASAN with this version.

## Performance

Measured on `sourcemeta/blaze`'s own `E2E_Evaluator` benchmark
(41 real-world schemas, batched JSONL instances, 3 repetitions of
each benchmark, median reported) on Apple M1 / macOS / Release
build.

| Aggregate | Δ vs baseline |
| --- | --- |
| Total wall time across all 41 schemas | **-8.80 %** |
| Mean per-schema delta | -8.87 % |
| Median per-schema delta | -9.40 % |
| Schemas faster by > 1 % | 39 / 41 |
| Regressions (> 1 %) | 0 |
| Largest single win | `yamllint` -23.35 % |
| Worst case | `jsconfig` +0.57 % (within noise) |

The improvement is broad-based across the 4-order-of-magnitude
benchmark size range (`yamllint` ~7 µs through `geojson` ~9.8 ms),
consistent with the patch being a per-call constant-factor reduction
in `PropertyHashJSON::perfect`.

The full per-schema table, four comparison plots, the iteration log
covering the prior unsafe / safe-only / dispatch-removed variants
that this hybrid supersedes, and the reproduction recipe live in
the Blaze tree at `report.md` and `benchmarks_results/`.

## What was tested locally

| Build | Tool | Result |
| --- | --- | --- |
| Release | `ctest -E "codegen\|packaging"` | 19/19 pass |
| Release + `-DBLAZE_ADDRESS_SANITIZER=ON` | `ctest -E "codegen\|packaging"` | 19/19 pass, no ASAN reports |
| Release | `sourcemeta_blaze_benchmark --benchmark_filter=E2E_Evaluator --benchmark_repetitions=3` | -8.80 % aggregate |

## Test plan

- [ ] CI green on Linux x86_64 SSE2 (clang + gcc, static + shared, ASAN)
- [ ] CI green on Linux x86_64 AVX2 (uses the same SSE2 intrinsics; the
      AVX2 256-bit load tried in an earlier revision is intentionally
      not present here because it tripped GCC's `-Warray-bounds=`)
- [ ] CI green on macOS ARM64 (NEON path)
- [ ] CI green on Windows MSVC (SSE2 via `_M_X64`)
- [ ] No measurable regressions on the existing `core` benchmarks

### DIFF
--- a/src/core/json/include/sourcemeta/core/json_hash.h
+++ b/src/core/json/include/sourcemeta/core/json_hash.h
@@ -7,6 +7,29 @@
 #include <cstring>    // std::memcpy
 #include <functional> // std::reference_wrapper
 
+// -----------------------------------------------------------------------------
+// SIMD backend auto-detection for `PropertyHashJSON::perfect`.
+//
+// We use SIMD *only* for the size >= 16 case, where the caller has guaranteed
+// at least 16 contiguous readable bytes at `data`. For size < 16 we fall back
+// to scalar `std::memcpy`, which the compiler already inlines to optimal
+// fixed-size moves for these small sizes and is, crucially, free of any
+// out-of-bounds read.
+// -----------------------------------------------------------------------------
+#if defined(__ARM_NEON) || defined(__ARM_NEON__)
+// NOLINTBEGIN(clang-diagnostic-error)
+// Some older clang-tidy versions choke when parsing newer Xcode/LLVM
+// `arm_neon.h` (e.g. unrecognized `__builtin_neon_*` symbols for bf16 and
+// complex-vector intrinsics). The header is correct, the diagnostic is a
+// clang-tidy bug; suppress it locally.
+#include <arm_neon.h>
+// NOLINTEND(clang-diagnostic-error)
+#define SOURCEMETA_HASH_SIMD_NEON 1
+#elif defined(__SSE2__) || defined(_M_X64) || defined(_M_AMD64)
+#include <emmintrin.h>
+#define SOURCEMETA_HASH_SIMD_SSE2 1
+#endif
+
 namespace sourcemeta::core {
 
 /// @ingroup json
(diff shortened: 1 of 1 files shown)


SouptikH:
Closing: end-to-end re-benchmark of the safe variant (no out-of-bounds read for size<16) shows only -3% aggregate on the 41-schema Blaze E2E_Evaluator suite, with regressions on flagship schemas (openapi +5.48%, cmake_presets +9.23%). The earlier -9.59% claim was measured with an unsafe over-read variant that ASAN correctly flagged. The remaining safe win is not worth shipping at the cost of those regressions. Will revisit with a different optimization target (likely StringSet::contains SIMD scan or schema-compiler peephole fusion).

Thanks for the CI feedback that surfaced the ASAN/GCC issues quickly.

SouptikH:
Reopened with the v5 *hybrid threshold* variant on commit `d2266e0`:

- size 1..7  → original scalar `memcpy` (compile-time-known size via the preserved 31-case switch)
- size 8..15 → SIMD via a 16-byte zero-padded stack bounce buffer (`memcpy(buf, src, size)` then `vld1q_u8(buf)`)
- size 16..31 → direct SIMD with overlapping tail load (caller guarantees ≥16 readable bytes)
- size ≥ 32 → unchanged collision-tag fallback

ASAN-safe: built locally with `-DBLAZE_ADDRESS_SANITIZER=ON`, full Blaze test suite 19/19 passes, including the three tests (`core.json`, `core.jsonpointer`, `core.uritemplate`) that flagged the previous variant.

Performance on the `E2E_Evaluator` suite (41 schemas, 3 reps, M1 Release): aggregate **-8.80 %**, 39/41 faster, 0 regressions.

CI issues from the previous attempt addressed:
- AVX2 256-bit load removed → fixes GCC `-Warray-bounds=`
- `prefix_mask_table` removed → no more `modernize-avoid-c-arrays` clang-tidy error
- `NOLINT` around `arm_neon.h` → works around older clang-tidy bug parsing Xcode 16's bf16/complex-vector intrinsics
- No path reads past the input → ASAN-clean

PR #2516, opened 2026-06-16
test(email): cover internationalized idn-email edge cases

vtushar06:
Core's `is_idn_email` already has good coverage, but a handful of RFC 6531 and RFC 5321 edge cases around the local-part and the domain delegation are not exercised yet. This adds them to `test/email/idn_email_test.cc`. I ran every input through `is_idn_email` (and `is_email`) on the current build before adding it, so each assertion matches the behavior as it stands.

The cases:

- The local-part vs domain normalization difference. RFC 6531 puts no NFC requirement on the local-part, while RFC 5890 section 2.3.2.1 requires a domain U-label to be in NFC. So the same decomposed "cafe" + U+0301 is a valid local-part but an invalid domain. The same split shows up for a leading combining mark, and for codepoints that are fine in a local-part but DISALLOWED in a domain (a symbol, a right-to-left character).
- Quoted-string local-parts: an empty quoted string, an "@" inside the quotes (ordinary content, not a second separator), and consecutive dots inside the quotes (the Dot-string rules do not apply there).
- The domain delegation to `is_idn_hostname`: a Bidi-rule violation, a DISALLOWED zero-width codepoint, and a leading combining mark are all rejected through the email path.
- Two mixed forms: a quoted non-ASCII local-part with a U-label domain, and a U-label local-part with an A-label domain.

Each test asserts both `is_idn_email` and `is_email`, which keeps the RFC 5321 vs RFC 6531 difference visible - a non-ASCII local-part is valid under `is_idn_email` but not `is_email`. Inputs are byte-escaped with the codepoints noted in a comment, and every test cites the governing RFC rule.

These are per-variant cases - the kind that fit in Core rather than the JSON Schema Test Suite, where one representative per mechanism is enough.

<a href="https://cubic.dev/pr/sourcemeta/core/pull/2516?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/email/idn_email_test.cc
+++ b/test/email/idn_email_test.cc
@@ -397,3 +397,135 @@ TEST(IdnEmail, invalid_domain_total_too_long) {
   EXPECT_FALSE(sourcemeta::core::is_idn_email("a@" + domain));
   EXPECT_FALSE(sourcemeta::core::is_email("a@" + domain));
 }
+
+// RFC 6531 §3.3 extends the local-part with UTF8-non-ascii but, unlike the
+// domain U-label (RFC 5890 §2.3.2.1), imposes no NFC requirement on it. So
+// "cafe" + U+0301 COMBINING ACUTE (NFD, bytes 63 61 66 65 CC 81) is a valid
+// local-part even though it is not in NFC.
+TEST(IdnEmail, valid_local_non_nfc) {
+  EXPECT_TRUE(sourcemeta::core::is_idn_email("cafe\xcc\x81@example.com"));
+  EXPECT_FALSE(sourcemeta::core::is_email("cafe\xcc\x81@example.com"));
+}
+
+// RFC 5890 §2.3.2.1: a domain U-label MUST be in NFC. The same "cafe" +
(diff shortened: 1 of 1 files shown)


vtushar06:
On the Copilot comments about `\xce`: that is a valid hex escape - `\x` takes hex digits and they are case-insensitive, so `\xce` is the same byte as `\xCE` (0xCE). The rest of this file already uses lowercase `\x..` escapes (the existing `\xce\xb1` U+03B1 tests, etc.), and it compiles, so I kept the new tests consistent with that. Each escape is bounded by the next `\` or a non-hex character, so the bytes are not ambiguous.

vtushar06:
@jviotti any review on this?

jviotti (approved):
Looks great to me! Thanks for the extra coverage and the excellent per-test comments

PR #2518, opened 2026-06-16
test(uri): expand is_iri coverage for ucschar boundaries, iprivate positions, and host forms

vtushar06:
I added 31 tests to `test/uri/uri_is_iri_test.cc` for parts of the RFC 3987 grammar the existing `is_iri` tests do not reach. Core already passes all of them, so this is coverage, not a fix.

What is new:

- ucschar range edges the current tests skip - the points just before and after the surrogate gap (U+D7FF) and the private-use gap (U+F900), the upper BMP bound (U+FFEF), the noncharacter blocks (U+FDD0-FDEF, U+FFFE, U+FFFF), the replacement character U+FFFD, and the plane-14 cutover at U+E0FFF / U+E1000.
- astral (non-BMP) ucschar, which the file did not cover at all - U+10000, U+1FFFD, U+20000, and the astral noncharacter U+1FFFE.
- the iprivate positional rule finished off - the existing tests have iprivate in query, path, fragment and host, so I added userinfo and port (both rejected) and the two supplementary private-use planes (U+F0000 and U+100000, valid only in the query).
- host and path forms the file was missing - an IPv4 host, IPvFuture in lower and upper case plus a non-hex version that is rejected, an opaque scheme (urn:), an absolute-path scheme (file:/), an explicit port, and the empty-hierarchy forms http: and http://.

Every test keeps the existing four-way checks - is_iri / is_iri_reference / is_uri / is_uri_reference - so each input also records the URI vs IRI split (a non-ASCII IRI is not a URI) and the absolute vs reference split.

How I checked the expected values: I ran each input through the four functions with a small driver linked against the library, confirmed each verdict against the RFC 3987 grammar, and wrote the tests from that. The uri test target builds and all 65 URI_is_iri tests pass.

RFC references:

- ucschar and iprivate ranges, and iprivate being allowed only in iquery - RFC 3987 section 2.2.
- IPvFuture and the case-insensitive host - RFC 3986 section 3.2.2, carried into RFC 3987 section 2.2 unchanged.

<a href="https://cubic.dev/pr/sourcemeta/core/pull/2518?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/uri/uri_is_iri_test.cc
+++ b/test/uri/uri_is_iri_test.cc
@@ -283,3 +283,251 @@ TEST(URI_is_iri, fragment_with_backslash_rejected) {
   EXPECT_FALSE(sourcemeta::core::URI::is_uri(input));
   EXPECT_FALSE(sourcemeta::core::URI::is_uri_reference(input));
 }
+
+TEST(URI_is_iri, ucschar_before_surrogate_gap_in_path) {
+  const std::string input{"https://example.com/\xED\x9F\xBF"};
+  EXPECT_TRUE(sourcemeta::core::URI::is_iri(input));
+  EXPECT_TRUE(sourcemeta::core::URI::is_iri_reference(input));
+  EXPECT_FALSE(sourcemeta::core::URI::is_uri(input));
+  EXPECT_FALSE(sourcemeta::core::URI::is_uri_reference(input));
+}
+
+TEST(URI_is_iri, ucschar_after_private_use_gap_in_path) {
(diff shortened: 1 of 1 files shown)


jviotti (approved):
Awesome!

PR #2626, opened 2026-07-10
test(time): expand is_rfc3339_duration coverage for ordering, week exclusivity, sign, fraction, and boundary cases

vtushar06:
I added 207 tests to `test/time/rfc3339_duration_test.cc` for parts of the RFC 3339 Appendix A grammar the existing `is_rfc3339_duration` tests do not reach. Core already passes all of them, so this is coverage, not a fix.

What is new:

- every date x time shape - the six date heads (D, MD, M, YMD, YM, Y) combined with the six time tails (H, HM, HMS, M, MS, S), which the existing file only partially covered.
- week exclusivity beyond the plain case - a week combined with a time block (`P1WT1H`) and a week after a zero-valued component (`P0Y1W`), which some implementations accept because they check exclusivity by value rather than syntax.
- the `1*DIGIT` rejections at each position - leading and internal sign (`-P1D`, `P-1D`), fraction with dot and comma (`P0.5Y`, `P0,5Y`), and scientific-notation exponent (`P1e2D`), including the huge-exponent form that other implementations mishandle.
- a dangling number immediately before `T` (`P1D2T3H`), where the number has no unit designator.
- component order and repeats (`P1M1Y`, `PT1M1H`, `P1Y2Y`), cross-placement (a time designator in the date part), and the `T`-edge cases.
- boundary values (zero, leading zero, large counts - `1*DIGIT` is unbounded) and BOM / NUL / control bytes. NUL inputs use an explicit-length `std::string_view` so the value is not truncated at the NUL.

How I checked the expected values: I ran each input through `is_rfc3339_duration` with a small driver linked against the library, confirmed each verdict against the RFC 3339 Appendix A grammar, and wrote the tests from that. The time test target builds and all 267 rfc3339_duration tests pass.

RFC references:

- the duration ABNF (P prefix, fixed component order, 1*DIGIT amounts, exclusive week) - RFC 3339 Appendix A.


test/time/rfc3339_duration_test.cc
```diff
@@ -1,242 +1,945 @@
...
+TEST(Time_rfc3339_duration, suite_valid_four_years) {
   EXPECT_TRUE(sourcemeta::core::is_rfc3339_duration("P4Y"));
 }
 
-TEST(suite_valid_two_weeks) {
+TEST(Time_rfc3339_duration, suite_valid_two_weeks) {
   EXPECT_TRUE(sourcemeta::core::is_rfc3339_duration("P2W"));
 }
+
+// Exhaustive RFC 3339 Appendix A duration coverage (per-mechanism variants).
```
jviotti: Can we avoid these section comments? It is easy to then add more tests that don't belong to them and thus making the comments non sense

test/time/rfc3339_duration_test.cc
```diff
@@ -1,242 +1,945 @@
...
+// Each verdict matches sourcemeta::core::is_rfc3339_duration as built.
+
+// valid structural shapes (date x time combinations)
+TEST(Time_rfc3339_duration, ext_valid_day_t_hour) {
+  EXPECT_TRUE(sourcemeta::core::is_rfc3339_duration("P1DT4H"));
+}
```
jviotti, line 246: If you see the other test files, we always have a blank line in between tests

jviotti (changes requested):

vtushar06:
@jviotti addressed both comments and CI is green now - ready for another look whenever you get a chance.

jviotti (approved):

PR #2644, opened 2026-07-21
test(crypto): expand is_uuid_like coverage for nibbles, Max, and non-hex characters

vtushar06:
I added 8 test blocks to `is_uuid_like`. Core already passes all of them, so this is coverage, not a fix.

What is new:

- `valid_max` - the Max UUID (all `f`), RFC 9562 section 5.10, alongside the existing Nil case.
- `valid_all_version_nibbles` and `valid_all_variant_nibbles` - loops over `0`-`f` at index 14 and index 19, pinning that `is_uuid_like` treats the version and variant as semantic fields, not syntactic constraints (RFC 9562 sections 4.1 and 4.2).
- `invalid_underscore_separator`, `invalid_leading_plus`, `invalid_hex_radix_prefix`, `invalid_non_ascii_digit` - characters that a lenient integer parser tolerates (`_`, `+`, `0x`, a Unicode decimal digit) but `HEXDIG` does not. `is_uuid_like` rejects them through its byte-level hex check, unlike a checker that validates the hex content with an integer parser.
- `invalid_trailing_newline` - a complete UUID followed by `\n`.

How I checked the expected values: I compiled and ran every block against the real `is_uuid_like` before staging - all 8 pass.

RFC references: RFC 9562 sections 4, 4.1, 4.2, 5.9, 5.10.

<a href="https://cubic.dev/pr/sourcemeta/core/pull/2644?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/crypto/crypto_uuid_test.cc
+++ b/test/crypto/crypto_uuid_test.cc
@@ -133,3 +133,199 @@ TEST(invalid_urn_prefix) {
   EXPECT_FALSE(sourcemeta::core::is_uuid_like(
       "urn:uuid:2eb8aa08-aa98-11ea-b4aa-73b441d16380"));
 }
+
+TEST(valid_max) {
+  EXPECT_TRUE(
+      sourcemeta::core::is_uuid_like("ffffffff-ffff-ffff-ffff-ffffffffffff"));
+}
+
+// RFC 9562 Section 4.2: the version nibble (index 14) is a semantic field, not
+// a syntactic constraint - every nibble 0-f is a valid string representation.
+// RFC 9562 Section 4.1: the variant nibble (index 19) is likewise
+// unconstrained.
+TEST(valid_version_nibble_0) {
+  EXPECT_TRUE(
+      sourcemeta::core::is_uuid_like("2eb8aa08-aa98-01ea-b4aa-73b441d16380"));
+}
+
+TEST(valid_version_nibble_1) {
+  EXPECT_TRUE(
+      sourcemeta::core::is_uuid_like("2eb8aa08-aa98-11ea-b4aa-73b441d16380"));
+}
+
+TEST(valid_version_nibble_2) {
+  EXPECT_TRUE(
+      sourcemeta::core::is_uuid_like("2eb8aa08-aa98-21ea-b4aa-73b441d16380"));
+}
+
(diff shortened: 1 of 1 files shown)


jviotti (approved):
Looks good but formatting and DCO are failing!

test/crypto/crypto_uuid_test.cc
```diff
@@ -133,3 +136,57 @@ TEST(invalid_urn_prefix) {
...
+    uuid[14] = nibble;
+    EXPECT_TRUE(sourcemeta::core::is_uuid_like(uuid));
+  }
+}
+
+// RFC 9562 Section 4.1: the variant nibble (index 19) is likewise semantic, so a
+// non-RFC-4122 variant (NCS 0-7, Microsoft c-d, future e-f) is still valid.
+TEST(valid_all_variant_nibbles) {
+  static constexpr std::string_view digits{"0123456789abcdef"};
+  for (const auto nibble : digits) {
```
jviotti: Actually, can we split this into multiple tests rather than looping inside?

test/crypto/crypto_uuid_test.cc
```diff
@@ -133,3 +136,57 @@ TEST(invalid_urn_prefix) {
...
+TEST(valid_max) {
+  EXPECT_TRUE(
+      sourcemeta::core::is_uuid_like("ffffffff-ffff-ffff-ffff-ffffffffffff"));
+}
+
+// RFC 9562 Section 4.2: the version nibble (index 14) is a semantic field, not a
+// syntactic constraint, so every nibble 0-f is a valid string representation.
+TEST(valid_all_version_nibbles) {
+  static constexpr std::string_view digits{"0123456789abcdef"};
+  for (const auto nibble : digits) {
```
jviotti: Same here with regards to the loop

vtushar06:
Hey @jviotti, Pr is ready for another look.

jviotti (approved):

PR #2646, opened 2026-07-21
test(jsonpointer): expand is_pointer coverage for the RFC 6901 section 3 grammar

vtushar06:
I added 120 tests for `sourcemeta::core::is_pointer` walking the full [RFC 6901 section 3](https://www.rfc-editor.org/rfc/rfc6901#section-3) string grammar directly. Core already passes all 120, so this is coverage, not a fix.

## Changes

The valid cases go into `test/jsonpointer/jsonpointer_parse_test.cc` and the invalid ones into `test/jsonpointer/jsonpointer_parse_error_test.cc`, next to the existing `is_pointer` checks (no new file).

## What is new

The tests cover every branch of the section-3 grammar, in both directions where a rule has a reject side:

- empty pointer (whole document) and runs of empty reference tokens (`/`, `//`, `///`, `////`, `/foo//bar`, `/foo/bar/`)
- single, multi, and deep token nesting
- every `~0` / `~1` escape combination (`~0`, `~1`, `~0~1`, `~1~0`, `~00`, `~11`, `~01`, `~10`, runs, escape-then-fraction)
- every invalid-escape shape (`~` at end of string, `~2`..`~9`, `~a`, `~-1`, `~~`, `~/`, dangling `~` interior and trailing)
- the leading-slash rule (a non-empty pointer that does not start with `/` is invalid: `a`, `0`, `a/a`, leading space/tab/newline)
- the section-6 `#`-fragment form as invalid section-3 string input (`#`, `#/`, `#a`, `#/foo`, `#/foo/0`, `#/a~1b`)
- the three `unescaped` sub-ranges pinned at their boundaries: `%x2E` (`.`, last before `/`), `%x30` (`0`, first after `/`), `%x7D` (`}`, last before `~`), `%x7F` (DEL, first of the third range), `%x80`, Latin-1, and astral code points
- index-looking tokens as valid strings (`/00`, `/01`, `/007`, `/1e0`, `/-1`, `/-`, huge numeric) - the section-4 leading-zero ban is array-index resolution semantics, not section-3 string validity
- literal `%` and `%2F` / `%7E` (no percent-decoding in the string form)
- trailing `\n` / `\r` / `\r\n` as valid (LF and CR are legal `unescaped` chars, `%x0A` / `%x0D`)
- NUL handling: an embedded NUL in a valid position stays valid (`/foo\0bar`), but invalidity placed after a NUL is still invalid (`/foo\0~`, `\0/foo`) - these use an explicit-length `std::string_view` so the NUL is not truncated
- normalization resistance: a fullwidth tilde `U+FF5E`, solidus `U+FF0F`, or digit `U+FF10` is an ordinary code point, not the structural `~` / `/` / an escape completer - RFC 6901 does no normalization

I transcribed the RFC 6901 section 3 ABNF into an oracle, cross-checked it against the published JSON Schema Test Suite `json-pointer.json` (40 tests, 0 disagreement), then ran all 120 inputs through the real `is_pointer` (a small driver linking `libsourcemeta_core_jsonpointer`) - 0 mismatch. The full jsonpointer unit suite passes: 739 tests, 0 failures.

## RFC references

- [RFC 6901 section 3](https://www.rfc-editor.org/rfc/rfc6901#section-3) - the string-representation grammar
- [RFC 6901 section 4](https://www.rfc-editor.org/rfc/rfc6901#section-4) - array-index leading-zero rule (resolution, not string validity)
- [RFC 6901 section 6](https://www.rfc-editor.org/rfc/rfc6901#section-6) - the separate URI-fragment representation


test/jsonpointer/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ sourcemeta_test(NAMESPACE sourcemeta PROJECT core NAME jsonpointer
     jsonpointer_empty_pointer_test.cc
     jsonpointer_get_test.cc
     jsonpointer_is_relative_pointer_test.cc
+    jsonpointer_is_pointer_test.cc
```
jviotti: We have `jsonpointer/jsonpointer_parse_test.cc` and `jsonpointer/jsonpointer_parse_error_test.cc` testing `is_pointer`. Can you move your new tests to those existing files?

vtushar06: Done - moved the valid cases into jsonpointer_parse_test.cc and the invalid ones into jsonpointer_parse_error_test.cc, and dropped the extra file. Full jsonpointer unit suite passes (739 tests).

test/jsonpointer/jsonpointer_parse_error_test.cc
```diff
@@ -74,3 +75,143 @@ TEST(foo_tilde_tilde) {
   EXPECT_POINTER_PARSE_ERROR(input, 6);
   EXPECT_FALSE(sourcemeta::core::is_pointer(input));
 }
+
+// is_pointer coverage for the RFC 6901 section 3 string grammar
```
jviotti: Let's get rid of section comments!

test/jsonpointer/jsonpointer_parse_test.cc
```diff
@@ -714,3 +714,260 @@ TEST(to_pointer_default_string_view) {
   const auto pointer = sourcemeta::core::to_pointer(input);
   EXPECT_EQ(pointer.size(), 0);
 }
+
+// is_pointer coverage for the RFC 6901 section 3 string grammar
```
jviotti: Same here

jviotti:
Dropped some more comments. Also note that Windows builds on CI seem to be failing, in case you want to have a shot at debugging those

vtushar06:
@jviotti removed the section comments - ready for another look, thanks!

test/jsonpointer/jsonpointer_parse_test.cc
jviotti, line 684: Note that other tests in this file like this one do more than just a `EXPECT_TRUE` check. The check the size and individual components. Can we do the same for extra coverage and uniformity?

test/jsonpointer/jsonpointer_parse_error_test.cc
```diff
@@ -74,3 +75,139 @@ TEST(foo_tilde_tilde) {
   EXPECT_POINTER_PARSE_ERROR(input, 6);
```
jviotti, line 75: Similarly, in this file see that existing tests don't just do the `EXPECT_FALSE` but also a `EXPECT_POINTER_PARSE_ERROR` to check parsing itself triggers an issue, at the specific corresponding column.

Can we imitate the style for consistency and extra coverage?

test/jsonpointer/jsonpointer_parse_test.cc
vtushar06: Done - the valid cases now also check to_pointer's size and each token's kind and value, matching the existing tests.

test/jsonpointer/jsonpointer_parse_error_test.cc
vtushar06: Done - the invalid cases now check EXPECT_POINTER_PARSE_ERROR at the corresponding column too. Also dropped the duplicate cases and fixed the MSVC C4125 Windows warning (split an octal escape that a digit followed).

vtushar06:
@jviotti addressed both - matched the existing test style (size/component checks for valid, parse-error column for invalid), dropped the duplicates, and fixed the Windows C4125 warning. Ready for another look, thanks!

test/jsonpointer/jsonpointer_parse_test.cc
vtushar06: yes sure

test/jsonpointer/jsonpointer_parse_error_test.cc
vtushar06: done

jviotti (approved):

PR #2683, opened 2026-07-25
test(dns): expand is_hostname coverage for length, hyphen, dot, and A-label branches

vtushar06:
I added 94 tests for `is_hostname`. Core already passes all of them, so this is coverage, not a fix.

I went through `src/core/dns/hostname.cc` branch by branch and wrote tests for the ones the existing 66 did not reach.

## What is new

- **Total length.** The existing cap tests reach 255 through four 63-byte labels. These reach it through 128 single-character labels instead, so the arithmetic is exercised rather than one shape. Plus 63/64 in both a middle and a last label, which are different code paths.
- **Hyphen placement.** Leading, trailing, lone `-`, consecutive hyphens, and a 63-byte label that is almost all hyphens.
- **Dot structure.** Empty labels, leading and trailing dots, repeated dots.
- **The `xn--` A-label branch**, including the case-insensitive prefix match (`Xn--`, `xN--`, `XN--`).
- **Line terminators, whitespace and embedded NUL.** The NUL cases use an explicit-length `std::string_view(..., N)` - written as a plain C string literal they would truncate at the NUL and silently test a different input.
- **Non-ASCII UTF-8** in a label.
- **Three loops** sweeping all 256 byte values in leading, interior and trailing position, so the accepted character set is pinned by construction rather than by sampling.

Every input was run through the real `is_hostname` before I wrote the expectation down, so none of these are hand-guessed. The file compiles clean and the dns target reports **277 passed, 0 failed**. `clang-format` is idempotent on it.

## One deliberate inconsistency

The existing tests in this file assert `is_idn_hostname` alongside `is_hostname`. Mine assert `is_hostname` only. A few of these inputs are ones where the two predicates currently disagree, so pairing them here would have written that disagreement in as expected behaviour. I would rather raise it on its own, which I will do separately.

## RFC references

- [RFC 1123 section 2.1](https://www.rfc-editor.org/rfc/rfc1123#section-2.1) - host name syntax, and the first-character relaxation
- [RFC 952](https://www.rfc-editor.org/rfc/rfc952) - the `<name>` grammar and its character repertoire
- [RFC 1035 section 2.3.1](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1) and [section 3.1](https://www.rfc-editor.org/rfc/rfc1035#section-3.1) - the letter/digit sets and the length limits

<a href="https://cubic.dev/pr/sourcemeta/core/pull/2683?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/dns/hostname_test.cc
+++ b/test/dns/hostname_test.cc
@@ -450,3 +450,588 @@ TEST(invalid_at_sign) {
   EXPECT_FALSE(sourcemeta::core::is_hostname("user@host"));
   EXPECT_FALSE(sourcemeta::core::is_idn_hostname("user@host"));
 }
+
+// Total length cap. RFC 1123 §2.1: "Host software MUST handle host
+// names of up to 63 characters and SHOULD handle host names of up to 255
+// characters." The existing tests only reach the cap through four 63-byte
+// labels, so these reach it through many short ones instead
+
+TEST(valid_total_255_many_short_labels) {
+  // 128 single-character labels and 127 dots is 255 bytes, right at the cap
+  std::string hostname{"a"};
+  for (int index = 0; index < 127; ++index) {
+    hostname.append(".a");
+  }
+  EXPECT_EQ(hostname.size(), 255u);
+  EXPECT_TRUE(sourcemeta::core::is_hostname(hostname));
+}
+
+TEST(invalid_total_256_many_short_labels) {
+  // the same shape one byte over the cap, with every label still well formed
+  std::string hostname{"aa"};
+  for (int index = 0; index < 127; ++index) {
+    hostname.append(".a");
+  }
+  EXPECT_EQ(hostname.size(), 256u);
+  EXPECT_FALSE(sourcemeta::core::is_hostname(hostname));
+}
+
+TEST(invalid_total_far_over_cap) {
+  // the total-length check runs before any label scan, so an input far past
+  // the cap is rejected without walking it
+  EXPECT_FALSE(sourcemeta::core::is_hostname(std::string(100000, 'a')));
+}
+
+// Per-label cap in non-leading positions. RFC 1035 §2.3.4 caps a label at 63
+// octets, and the existing tests only exercise the first label
+
+TEST(valid_middle_label_63) {
+  // 63-character label between two other labels
+  EXPECT_TRUE(
+      sourcemeta::core::is_hostname("a." + std::string(63, 'a') + ".b"));
+}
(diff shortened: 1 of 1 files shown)


test/dns/hostname_test.cc
```diff
@@ -450,3 +450,588 @@ TEST(invalid_at_sign) {
   EXPECT_FALSE(sourcemeta::core::is_hostname("user@host"));
   EXPECT_FALSE(sourcemeta::core::is_idn_hostname("user@host"));
 }
+
+// Total length cap. RFC 1123 §2.1: "Host software MUST handle host
```
jviotti: Can we avoid the section comments?

test/dns/hostname_test.cc
```diff
@@ -450,3 +450,588 @@ TEST(invalid_at_sign) {
...
+      std::string_view("example.com\0.evil.test", 22)));
+}
+
+// The label alphabet, swept over every byte value. RFC 952 §B builds
+// <name> out of <let>, <let-or-digit> and <let-or-digit-or-hyphen>, and
+// RFC 1123 §2.1 changes only the first-character rule, so the accepted set is
+// closed and can be checked exhaustively rather than sampled
+
+TEST(byte_alphabet_interior_position) {
+  for (int value = 0; value < 256; ++value) {
```
jviotti: Can we avoid loops in tests? If this is about those 256 characters, worst case you can have a single `TEST` that has literally 256 sequential `EXPECT_TRUE/FALSE` calls? Though better to split them into multiple `TEST`s at least into groups

test/dns/hostname_test.cc
vtushar06: Done - removed the section banners, back to one line per test.

test/dns/hostname_test.cc
vtushar06: Done - expanded into literal calls, grouped by ASCII range so no single test is huge.

vtushar06:
@jviotti I have addressed both - dropped the section banners for one line per test, and expanded the byte sweep into literal calls grouped by ASCII range.

jviotti (approved):

PR #2684, opened 2026-07-25
test(time): expand is_rfc3339_fulldate coverage for length, digit-slot, separator, and leap branches

vtushar06:
I added 265 tests to `test/time/rfc3339_fulldate_test.cc`. Core already passes all of them, so this is coverage, not a fix. I organised them by the branches in `src/core/time/rfc3339_fulldate.cc` rather than by input shape, so every `if` in the function has assertions on both sides of it.

I measured the gap instead of guessing at it. I built single-branch mutants of `is_rfc3339_fulldate` and checked which ones the existing 66 tests can distinguish from the real function. The existing file kills 3 of 10 usable mutants; with these additions it kills 10 of 10. The seven that only the new tests catch:

| mutant | what it breaks | why the existing tests miss it |
|---|---|---|
| `is_digit` also accepts any byte `>= 0x80` | `date-fullyear` digit check | every existing non-ASCII test is 12-14 bytes and is rejected by `value.size() != 10` before `is_digit` runs |
| `value[4] != '-'` never checked | first separator | `"2024/01/15"` changes *both* separators, so the second check still rejects it |
| `value[7] != '-'` never checked | second separator | same |
| `!is_digit(value[3])` dropped | last digit of the year | `"YYYY-01-01"` fails on `value[0]` first |
| `!is_digit(value[6])` dropped | second digit of the month | an alphabetic probe puts the month far out of range, so it is rejected anyway |
| `!is_digit(value[9])` dropped | second digit of the day | same |
| separator also accepts `.` `_` `/` | first separator | no test uses a single wrong separator in one position |

## What is new

- **length guard** (`value.size() != 10`) - lengths 1-7, 9, 11, 12, 13, 20 and 1000, plus every terminator variant (LF, CR, CRLF, TAB, NUL, NBSP, U+2028, BOM, zero-width space) in leading and trailing position.
- **a high byte that actually reaches `is_digit`** - inputs of exactly 10 bytes carrying one non-ASCII sequence, one per UTF-8 encoded length (2-byte Arabic-Indic, 3-byte Bengali and fullwidth, 4-byte mathematical bold), in the year, month, day and separator slots, plus a lone continuation byte, a lone lead byte, `0xFF`, and a C1 control byte. Every non-ASCII case in the file today is 12 or more bytes and never gets past the length guard.
- **each digit slot individually** - a non-digit at each of positions 0, 1, 2, 3, 5, 6, 8, 9, so no slot's check is load-bearing for another's.
- **the second digit of a 2DIGIT field landing back in range** - `':'`, `';'`, `'<'` after `date-month`'s first digit and `':'` through `'?'` after `date-mday`'s, since `':' - '0' == 10` puts the field back inside its legal range.
- **each separator position separately** - 15 wrong characters at `value[4]` alone, at `value[7]` alone, and at both.
- **`date-month` range** - 00, 01, 02, 09, 10, 11, 12, 13, 14, 19, 20, 29, 30, 50, 90, 99.
- **`date-mday` per month** - 00, 01, the section 5.7 maximum, maximum + 1, and 99 for all twelve months, plus February 28/29/30 in both a leap and a non-leap year.
- **`is_leap_year` branches** - February 29 across 39 years covering all four branches (not divisible by 4; by 4 not 100; by 100 not 400; by 400), including 0, 1, 100, 400, 1700, 1800, 1900, 2000, 2100, 2200, 2300, 2400, 9600, 9900 and 9996.
- **signs and extended-year forms** - `+`/`-` on a four-digit year, on a five-digit year, on `0000`, and `10000-01-01` unsigned.
- **range boundaries** - `0000-01-01`, `0001-01-01`, `0001-12-31`, `9998-12-31`, `9999-01-01`, `9999-02-28`.

I did not write the expected verdicts by hand. Every input was run through the real `is_rfc3339_fulldate`, compiled straight from this branch's source, and the `EXPECT_TRUE` / `EXPECT_FALSE` was emitted from what the function actually returned - then the whole file was compiled and run again to confirm. `sourcemeta_core_time_unit` builds clean and reports **693 passed, 0 failed**. `clang-format` is clean and the diff is additions only.

Two inputs are emitted as `std::string_view("...", N)` rather than a plain literal, because they contain an embedded NUL and a plain C string literal would truncate there - the test would silently check a different, shorter input.

I also cross-checked the verdicts against an independent transcription of the ABNF (three separate checkers written in different styles, agreeing with each other and with the published JSON Schema Test Suite on all 309 of its `date` string cases) so the assertions follow RFC 3339 rather than Core's current behaviour.

## RFC references

- [RFC 3339 section 5.6](https://www.rfc-editor.org/rfc/rfc3339#section-5.6) - `full-date = date-fullyear "-" date-month "-" date-mday`
- [RFC 3339 section 5.7](https://www.rfc-editor.org/rfc/rfc3339#section-5.7) - the maximum `date-mday` per month
- [RFC 3339 Appendix C](https://www.rfc-editor.org/rfc/rfc3339#appendix-C) - the leap year rule
- [RFC 5234 section 3.4](https://www.rfc-editor.org/rfc/rfc5234#section-3.4) - `DIGIT = %x30-39`

### DIFF
--- a/test/time/rfc3339_fulldate_test.cc
+++ b/test/time/rfc3339_fulldate_test.cc
@@ -266,3 +266,1114 @@ TEST(invalid_bengali_digit_in_year) {
 TEST(invalid_alpha_year) {
   EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("YYYY-01-01"));
 }
+
+TEST(length_01_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("2"));
+}
+
+TEST(length_02_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("22"));
+}
+
+TEST(length_03_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("222"));
+}
+
+TEST(length_04_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("2222"));
+}
+
+TEST(length_05_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("22222"));
+}
+
+TEST(length_06_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("222222"));
+}
+
+TEST(length_07_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("2222222"));
+}
+
+TEST(length_09_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("222222222"));
+}
+
+TEST(length_11_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("22222222222"));
+}
+
+TEST(length_12_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("222222222222"));
+}
+
+TEST(length_13_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("2222222222222"));
+}
+
+TEST(length_20_digits_only) {
+  EXPECT_FALSE(sourcemeta::core::is_rfc3339_fulldate("22222222222222222222"));
+}
+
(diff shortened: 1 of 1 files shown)


vtushar06:
hey @jviotti, let me know if any thing being left here :)

jviotti (approved):
I had probably missed this one! Thanks!

PR #2698, opened 2026-07-31
test(email): expand is_idn_email coverage for the address-literal branch

vtushar06:
I added 39 tests to `idn_email_test.cc` covering the address-literal branch of `is_idn_email`. Core already passes all of them, so this is coverage, not a fix.

What is new: the branch had almost nothing behind it. `is_general_address_literal` had no tests at all, and `is_snum`'s leading-zero behaviour - the thing that separates it from the RFC 3986 dec-octet behind `is_ipv4` - was not pinned anywhere, despite being called out in a comment.

- `Snum` (13 tests): leading zeros in each position, padded zero octets, the 255 maximum, one over, four digits, wrong octet counts, and the dot-placement errors.
- `General-address-literal` (9 tests): the `tag:content` form, the shortest valid form, a leading-hyphen tag, a trailing-hyphen tag, empty tag and empty content, and the `dcontent` exclusions for SP and `[`.
- IPv6 literals (5 tests): the fully compressed form, an uppercase `IPV6:` tag, `IPv6v4-comp`, `IPv6v4-full`, and the empty payload.
- Literal shape (5 tests): empty, unterminated, doubled brackets, a label after the closing bracket, and a trailing space inside.
- The `@` delimiter (2 tests): U+FF20 and U+FE6B, neither of which is the delimiter.
- Local-part code points (4 tests): a C1 control, two noncharacters, and a supplementary private-use code point - all admitted by `UTF8-non-ascii` with no property filter.
- Quoted-string structure (3 tests): unterminated, text after the closing DQUOTE, and a quoted-pair escaping the DQUOTE.

How I checked the expected values: every input was run against the built `is_idn_email` and `is_email` before being written, and each expectation is the RFC verdict rather than the observed one - the two agreed on all 43 candidates I probed. `sourcemeta_core_email_unit` is 393 passed / 0 failed with these added.

Two of my initial expectations were wrong and Core was right, which is worth recording. `user@[-tag:abc]` is valid, because `Ldh-str = *( ALPHA / DIGIT / "-" ) Let-dig` only constrains the last character, unlike `sub-domain = Let-dig [Ldh-str]`. And `user@[IPv4:192.0.2.1]` is a syntactically valid General-address-literal, since `IPv4` is not a tag RFC 5321 defines. Both are now tests with the reasoning in the comment.

RFC references:

- RFC 5321 section 4.1.3 (Address Literals - `Snum`, `dcontent`, `General-address-literal`, the IPv6 forms): https://www.rfc-editor.org/rfc/rfc5321#section-4.1.3
- RFC 5321 section 4.1.2 (`Mailbox`, `Ldh-str`, `Quoted-string`, `quoted-pairSMTP`): https://www.rfc-editor.org/rfc/rfc5321#section-4.1.2
- RFC 5234 section 2.3 (ABNF literal strings are case-insensitive, so the `IPv6:` tag matches in any case): https://www.rfc-editor.org/rfc/rfc5234#section-2.3
- RFC 6532 section 3.1 and RFC 3629 section 4 (`UTF8-non-ascii`, for the local-part code-point tests): https://www.rfc-editor.org/rfc/rfc6532#section-3.1

<a href="https://cubic.dev/pr/sourcemeta/core/pull/2698?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/email/idn_email_test.cc
+++ b/test/email/idn_email_test.cc
@@ -528,3 +528,256 @@ TEST(valid_ulabel_local_and_alabel_domain) {
   EXPECT_TRUE(sourcemeta::core::is_idn_email("\xce\xb1@xn--nxasmq6b.com"));
   EXPECT_FALSE(sourcemeta::core::is_email("\xce\xb1@xn--nxasmq6b.com"));
 }
+
+// --- RFC 5321 §4.1.3 address-literal: Snum ---------------------------------
+
+// RFC 5321 §4.1.3: Snum = 1*3DIGIT "representing a decimal integer value in
+// the range 0 through 255". The rule constrains the VALUE, not the digit
+// count, so leading zeros are inside the grammar. This is the documented
+// difference from the RFC 3986 dec-octet that backs is_ipv4, which forbids
+// them - so an implementation that reuses its ipv4 checker here is wrong.
+TEST(valid_address_literal_snum_leading_zero) {
+  EXPECT_TRUE(sourcemeta::core::is_idn_email("user@[01.0.0.1]"));
+  EXPECT_TRUE(sourcemeta::core::is_email("user@[01.0.0.1]"));
+}
+
+TEST(valid_address_literal_snum_leading_zero_last_octet) {
+  EXPECT_TRUE(sourcemeta::core::is_idn_email("user@[0.0.0.01]"));
+  EXPECT_TRUE(sourcemeta::core::is_email("user@[0.0.0.01]"));
+}
+
+TEST(valid_address_literal_snum_leading_zeros_every_octet) {
+  EXPECT_TRUE(sourcemeta::core::is_idn_email("user@[001.002.003.004]"));
+  EXPECT_TRUE(sourcemeta::core::is_email("user@[001.002.003.004]"));
+}
+
+TEST(valid_address_literal_snum_padded_zero_octets) {
+  EXPECT_TRUE(sourcemeta::core::is_idn_email("user@[000.000.000.000]"));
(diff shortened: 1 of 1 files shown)


test/email/idn_email_test.cc
```diff
@@ -528,3 +528,256 @@ TEST(valid_ulabel_local_and_alabel_domain) {
...
+  EXPECT_FALSE(sourcemeta::core::is_email("\xef\xb7\x90@example.com"));
+}
+
+// U+F0000, a supplementary private-use codepoint (F3 B0 80 80)
+TEST(valid_local_supplementary_private_use) {
+  EXPECT_TRUE(sourcemeta::core::is_idn_email("\xf3\xb0\x80\x80@example.com"));
+  EXPECT_FALSE(sourcemeta::core::is_email("\xf3\xb0\x80\x80@example.com"));
+}
+
+// --- Quoted-string structure -----------------------------------------------
```
jviotti: Can you remove section comments?

jviotti (changes requested):
Looks good, but there are still lots of section comments!

vtushar06:
i think it is ready for final look @jviotti

jviotti (approved):

PR #2726, opened 2026-08-14
test(dns): expand is_idn_hostname_uts46 coverage for the mapping step, Bidi and ContextO

vtushar06:
I added 25 tests to `test/dns/idn_hostname_uts46_test.cc`. Core already passes all of them, so this is coverage, not a fix.

The file had 19 tests, against 98 for the strict function, and the gap was mostly in the parts of UTS #46 that only the lookup profile reaches.

## What is new

- **The order of the processing steps.** 63 fullwidth `ａ` is 189 octets before mapping and 63 after, so the length limit has to be applied to the mapped form. `ｘｎ--nxasmq6b` only becomes an A-label after step 1, so the `xn--` test has to run after mapping. There are also the cases where the mapping manufactures a violation that the input does not have - `－abc` becomes a leading hyphen, `a．．b` becomes an empty label, `ab－－cd` becomes `--` in the third and fourth positions.
- **The mapping table's own status classes.** `mapped` reached through a few different routes (U+1D400, U+24D0, U+FB01), `ignored` via U+FE00, `deviation` via U+03C2 kept under nontransitional processing, and the NFC normalisation on its own.
- **The Bidi rule's conditions 2, 3 and 4.** The strict file covers condition 1 and the digit-first case; nothing covered an L inside an RTL label, an RTL label ending in ON, or a label mixing EN and AN.
- **CONTEXTO appendix A.9.** `IdnaTestV2.txt` states in its own header that "The CONTEXTO tests are optional for client software, and not tested here", so the suite file cannot cover this even though Core implements it. The extended Arabic-Indic digit rule and the per-label scope of A.8/A.9 are both exercised.

Every assertion is the live verdict, read off the real functions through a driver linked against `sourcemeta::core::dns` rather than written by hand. Each test asserts both `is_idn_hostname_uts46` and `is_idn_hostname`, so the pair also documents where the two profiles part company - 9 of the 25 differ between them.

Any input containing a byte a C string literal would mangle is emitted as an explicit-length `std::string_view` with a `static_assert` on the length, so a shortened literal fails to compile rather than silently testing something else.

`sourcemeta_core_dns_unit` builds and runs clean: 335 passed, 0 failed. clang-format applied, and the diff is additions only.

One case I deliberately left out: an A-label with an uppercase Punycode body. The strict function rejects it today, and asserting that would pin a verdict that RFC 5891 section 5.3 ("first ensuring that the A-label is entirely in lowercase"), RFC 5891 section 3.1 and RFC 3492 section 5 all point the other way. That seemed like the wrong thing to lock into a coverage PR, so I have raised it on the suite side instead.

## RFC references

- [UTS #46 section 4](https://www.unicode.org/reports/tr46/#Processing) and [section 4.1](https://www.unicode.org/reports/tr46/#Validity_Criteria)
- [RFC 5892 appendix A.8 and A.9](https://www.rfc-editor.org/rfc/rfc5892#appendix-A.8)
- [RFC 5893 section 2](https://www.rfc-editor.org/rfc/rfc5893#section-2)

<a href="https://cubic.dev/pr/sourcemeta/core/pull/2726?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/test/dns/idn_hostname_uts46_test.cc
+++ b/test/dns/idn_hostname_uts46_test.cc
@@ -1,6 +1,8 @@
 #include <sourcemeta/core/dns.h>
 #include <sourcemeta/core/test.h>
 
+#include <string_view> // std::string_view
+
 // A plain ASCII host name is unaffected by the mapping
 TEST(valid_ascii_dotted) {
   EXPECT_TRUE(sourcemeta::core::is_idn_hostname_uts46("www.example.com"));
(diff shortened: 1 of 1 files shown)


jviotti (approved):
Looks good! Keep in mind the new CLA CI new requirement, which is now a blocker for merging. Let me know when that's done!

PR #2790, opened 2026-09-06
feat: add ANSI terminal styling utilities

HarshPopat23:
## Summary
- Add a reusable Core module `terminal` (`src/lang/terminal`) for ANSI terminal styling.
- Automatically suppress escape codes when the output destination is not connected to an interactive terminal.
- Provide a thread-safe global disable policy (`ColorPolicy::Disabled`) for future `--no-color` CLI support.
- Support standardized ANSI/ECMA-48 SGR sequences for text styles (Bold, Reset) and foreground colors (Red, Green, Yellow, Blue, Cyan).
- Provide clean compatibility with modern formatted output (`std::format`, `std::print`, `std::println`).
- Support safe cross-platform behavior (POSIX `isatty`, Windows console VT processing via `SetConsoleMode`).
- Add comprehensive, deterministic unit tests independent of CI TTY allocation.

## Motivation
Sourcemeta CLIs (including the JSON Schema CLI and Sourcemeta One CLIs) need consistent terminal presentation without duplicating platform detection and ANSI handling. Keeping this functionality in Core allows applications to share one tested, polished implementation.

## Scope
This PR only introduces the reusable Core functionality. It does not change the JSON Schema CLI or add a `--no-color` command-line option.

## Testing
- Unit tests added in `test/terminal/terminal_test.cc` covering:
  - Exact SGR sequences for all supported styles and compound styles.
  - Reset sequences (`\033[0m`).
  - Style bitwise operators (`|`, `&`, `~`, `^`).
  - Enabled vs. suppressed styling.
  - Empty text handling (no dangling escape codes).
  - Global policy overrides (disabling styling suppresses all escape codes).
  - Independent stdout/stderr queries.
- Clean formatting verified with `git diff --check`.

## Platform notes
- **POSIX**: Uses `isatty` on `STDOUT_FILENO` and `STDERR_FILENO` to detect interactive terminals; redirects automatically suppress styling.
- **Windows**: Checks console handles via `GetStdHandle` and `GetConsoleMode`, enabling `ENABLE_VIRTUAL_TERMINAL_PROCESSING` where supported, and gracefully falling back to plain text when unsupported or redirected.

## Output Screenshot
image

<a href="https://cubic.dev/pr/sourcemeta/core/pull/2790?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+/// ```cpp
+/// #include <sourcemeta/core/terminal.h>
+/// ```
+
+namespace sourcemeta::core {
+
+/// @ingroup terminal
+///
+/// Standard output stream destinations for interactive terminal detection.
+enum class Stream : std::uint8_t {
```
jviotti: Can we prefix any types we add here with `Terminal`? Like `Stream` => `TerminalStream`?

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+///
+/// ```cpp
+/// #include <sourcemeta/core/terminal.h>
+/// ```
+
+namespace sourcemeta::core {
+
+/// @ingroup terminal
+///
+/// Standard output stream destinations for interactive terminal detection.
```
jviotti: You say here `output stream` but the type name is just `Stream`, which makes me think you should be modelling `Stdin` too?

jviotti: Also, I believe the POSIX standard defines these streams, so it would be good to link to the appropriate sections of that spec in the Oxygen for each of the enum members, etc?

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+  Stderr
+};
+
+/// @ingroup terminal
+///
+/// Global color policy governing ANSI styling output.
+enum class ColorPolicy : std::uint8_t {
+  /// Color is permitted when destination is an interactive terminal.
+  Automatic,
+  /// Styling is unconditionally suppressed (e.g. for `--no-color`).
```
jviotti: I wouldn't mention command line arguments here?

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+  /// Standard error stream (stderr).
+  Stderr
+};
+
+/// @ingroup terminal
+///
+/// Global color policy governing ANSI styling output.
+enum class ColorPolicy : std::uint8_t {
+  /// Color is permitted when destination is an interactive terminal.
+  Automatic,
```
jviotti: Maybe `WhenInteractive` is a better name given the description you gave it above? Also can you check if there is any standard, like POSIX or ANSI, mentioning this idea of avoiding colours if the terminal is non-interactive?

I wonder if some of those specs do standardise that, and in such case we can cite here and make sure we implement accordingly

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+  /// Styling is unconditionally suppressed (e.g. for `--no-color`).
+  Disabled
+};
+
+/// @ingroup terminal
+///
+/// Standard ANSI/ECMA-48 Select Graphic Rendition (SGR) text styles and
```
jviotti: Exactly. I love that you mention specific standards btw. Definitely do it as much as you can here, so its easy to check where these things come from, and for us to make sure we implement them well and don't get derailed to something non-standard

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+/// XOR two styles.
+[[nodiscard]] constexpr auto operator^(Style lhs, Style rhs) noexcept -> Style {
+  using Underlying = std::underlying_type_t<Style>;
+  return static_cast<Style>(static_cast<Underlying>(lhs) ^
+                            static_cast<Underlying>(rhs));
+}
+
+/// @ingroup terminal
+/// Reset style constant.
+inline constexpr Style RESET{Style::None};
```
jviotti: Beyond the `Terminal` prefix I suggested on types so we don't pollute the namespace, I think these ones are ANSI specific right? If so, try to have very specific concrete names for these things that directly derive from the conventions of the relevant standards. For example, maybe it is something like `TERMINAL_ANSI_SGR_ESCAPE_RESET` or whatever it is (check!)

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+
+/// @ingroup terminal
+/// Bold text style constant.
+inline constexpr Style BOLD{Style::Bold};
```
jviotti: But overall these seem like direct instantiations of the `enum class` above? Why have them though? Isn't the enum class already enough?

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+
+/// @ingroup terminal
+/// Cyan foreground color constant.
+inline constexpr Style CYAN{Style::Cyan};
+
+/// @ingroup terminal
+///
+/// Check whether the specified stream is connected to an interactive terminal.
+///
+/// This is a read-only check that does not alter console state. Returns `true`
```
jviotti: I think the whole `Returns true if ...., and false otherwise` is just over-explaining the obvious?

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+/// For example:
+///
+/// ```cpp
+/// #include <sourcemeta/core/terminal.h>
+///
+/// const bool is_stdout_interactive =
+///   sourcemeta::core::is_terminal(sourcemeta::core::Stream::Stdout);
+/// ```
+SOURCEMETA_CORE_TERMINAL_EXPORT
+auto is_terminal(Stream stream) noexcept -> bool;
```
jviotti: Maybe we should call this `terminal_is_interactive()`? Just to make it more obvious and lead with the module name

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+///
+/// For example:
+///
+/// ```cpp
+/// #include <sourcemeta/core/terminal.h>
+///
+/// sourcemeta::core::set_color_policy(sourcemeta::core::ColorPolicy::Disabled);
+/// ```
+SOURCEMETA_CORE_TERMINAL_EXPORT
+auto set_color_policy(ColorPolicy policy) noexcept -> void;
```
jviotti: Shouldn't we be setting this per stream instead of globally? Just asking as I'm not an expert on how these things should work.

Also, are there any potential issues with concurrency / multi-threading here?

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+///
+/// Retrieve the current global color policy.
+SOURCEMETA_CORE_TERMINAL_EXPORT
+auto color_policy() noexcept -> ColorPolicy;
+
+/// @ingroup terminal
+///
+/// Determine whether color/styling should be applied to the specified stream.
+///
+/// Returns `true` if and only if `color_policy()` is `ColorPolicy::Automatic`
```
jviotti: I think this comment is just explaining the internal implementation? Let's avoid such comments, as they can easily go out of date

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+/// ```cpp
+/// #include <sourcemeta/core/terminal.h>
+/// #include <cassert>
+///
+/// const auto styled = sourcemeta::core::paint("Notice",
+///   sourcemeta::core::BOLD | sourcemeta::core::GREEN, true);
+/// assert(styled == "\033[1;32mNotice\033[0m");
+/// ```
+SOURCEMETA_CORE_TERMINAL_EXPORT
+auto paint(std::string_view text, Style style, bool enabled) -> std::string;
```
jviotti: I understand how this is supposed to work, but note it takes a string view as input and returns a full string. And that full string means a copy + allocation of the input string. Is there any way to accomplish what we need to accomplish while keeping copies/allocations to the minimum, for performance?

src/lang/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,256 @@
...
+///
+/// ```cpp
+/// #include <sourcemeta/core/terminal.h>
+/// #include <cassert>
+///
+/// const auto code = sourcemeta::core::sgr_code(sourcemeta::core::RED);
+/// assert(code == "\033[31m");
+/// ```
+SOURCEMETA_CORE_TERMINAL_EXPORT
+auto sgr_code(Style style) -> std::string;
```
jviotti: Same here on allocations. I presume the possible styles are finite, so we could pre-compute the possible sequences and return string views to those?

jviotti: Also here let's definitely try to align on language with the actual upstream specifications. For example you say here "SGR code", but on the Doxygen you say "ANSI/ECMA-48 SGR sequence". Which is the standard name for this? 

Try to revise all of these symbols and make sure we ALWAYS use the standard names to avoid confusing. Terminology is key here!

src/lang/terminal/include/sourcemeta/core/terminal.h
jviotti, line 1: Something to check: I believe everything we are defining in here is related to specific standards. Both ANSI and probably the POSIX standard behind UNIX.

Can you confirm? Because if so, if we are just implementing a set of standards here, `src/core/terminal` might be a better place. As `src/core` is the container for modules that map to standards, and `src/lang` are C++ only things.

src/lang/terminal/terminal.cc
```diff
@@ -0,0 +1,166 @@
...
+auto color_enabled(Stream stream) noexcept -> bool {
+  if (color_policy() == ColorPolicy::Disabled) {
+    return false;
+  }
+  return is_terminal(stream);
+}
+
+#if defined(_WIN32) && !defined(__MSYS__) && !defined(__CYGWIN__)
+
+auto is_terminal(Stream stream) noexcept -> bool {
```
jviotti: I think functions like this are not getting exercised on the tests at all? Can we try to get more test coverage?

src/lang/terminal/terminal.cc
```diff
@@ -0,0 +1,166 @@
...
+}
+
+auto color_enabled(Stream stream) noexcept -> bool {
+  if (color_policy() == ColorPolicy::Disabled) {
+    return false;
+  }
+  return is_terminal(stream);
+}
+
+#if defined(_WIN32) && !defined(__MSYS__) && !defined(__CYGWIN__)
```
jviotti: I think that if there are big portions that are platform specific, you probably can extract them into separate `.cc` files that get conditionally pulled it by the `CMakeLists.txt` of the module. See i.e. `src/core/crypto` for an example. That can be cleaner to read

src/core/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,227 @@
...
+
+/// @ingroup terminal
+///
+/// Standard I/O streams defined by POSIX.1-2017 (<unistd.h>).
+///
+/// @see https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/unistd.h.html
+enum class TerminalStream : std::uint8_t {
+  /// Standard input stream (POSIX.1-2017 STDIN_FILENO, 0).
+  /// @see https://pubs.opengroup.org/onlinepubs/9699919799/functions/stdin.html
+  Stdin,
```
jviotti: According to POSIX, I believe they define standard input to be file descriptor 0, standard output to be file descriptor 1, etc. I think it is already the case by how you did it here, but I wonder if explicitly saying i.e. `= 0` and forcing the number locking would be cool, as then you could transform any other constant variable from the standard library into this enum pretty transparently?

Just a tiny addition. Nothing super important

src/core/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,227 @@
...
+  /// @see
+  /// https://pubs.opengroup.org/onlinepubs/9699919799/functions/isatty.html
+  WhenInteractive,
+  /// Styling is unconditionally suppressed.
+  Disabled
+};
+
+/// @ingroup terminal
+///
+/// Text styles and foreground colors defined by ECMA-48 (5th Edition, 1991),
```
jviotti, line 84: Super. I love these citations. Exactly what we need to make sure we do it right

jviotti: Also learning more about these things myself as I go through this

src/core/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,227 @@
...
+/// Section 8.3.117 "SGR - SELECT GRAPHIC RENDITION".
+///
+/// Bitwise operators allow combining `TerminalStyle::Bold` with a foreground
+/// color (for example `TerminalStyle::Bold | TerminalStyle::Red`). Only one
+/// foreground color may be active at a time. If multiple foreground colors are
```
jviotti, line 90: Very minor, but given this precedence rule and i.e. only one foreground color active, etc, I wonder if:

- It would make sense to have a simple `constexpr` helper function in this header to check if a terminal style is fully valid or "canonical" or whatever term they use in the spec (check?). It can return true if i.e. no 2 foreground colours are applied at once, etc or any other rule the spec says

- Then, in the logic in this module that takes `TerminalStyle`, we check for this using `assert()`. That means that during development, if we somehow pass an incorrect terminal style, it will complain, which can help with correctness.

jviotti: The precedence would still apply in production builds just fine, but at least this is a nice correctness check?

src/core/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,227 @@
...
+/// combined, the first matching color in declaration order (Red > Green >
+/// Yellow > Blue > Cyan) takes precedence.
+///
+/// @see
+/// https://ecma-international.org/publications-and-standards/standards/ecma-48/
+enum class TerminalStyle : std::uint8_t {
+  /// Normal display / no style (plain text).
+  None = 0,
+  /// Bold or increased intensity (ECMA-48 SGR parameter 1).
```
jviotti, line 110: Beautiful parenthesis citations 🚀

src/core/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,227 @@
...
+  /// Blue foreground color (ECMA-48 SGR parameter 34).
+  Blue = 1 << 4,
+  /// Cyan foreground color (ECMA-48 SGR parameter 36).
+  Cyan = 1 << 5
+};
+
+/// @ingroup terminal
+///
+/// Combine two styles using bitwise OR.
+[[nodiscard]] constexpr auto operator|(TerminalStyle lhs,
```
jviotti: On `lhs` / `rhs`, in the codebase I try to avoid shortened variables, just to increase readability. In this case, maybe better to say `left` / `right`. Same below: `val` => `value`? It always reads better

src/core/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,227 @@
...
+                                       TerminalStyle rhs) noexcept
+    -> TerminalStyle {
+  using Underlying = std::underlying_type_t<TerminalStyle>;
+  return static_cast<TerminalStyle>(static_cast<Underlying>(lhs) ^
+                                    static_cast<Underlying>(rhs));
+}
+
+/// @ingroup terminal
+///
+/// Check whether the specified stream is connected to an interactive terminal.
```
jviotti: Note that in other modules beyond `src/core/terminal`, for functions/classes we tend to have a `For example: <C++ code block>`. Can you follow the same convention in this module?

src/core/terminal/terminal.cc
```diff
@@ -0,0 +1,157 @@
+#include "terminal_internal.h"
+#include <sourcemeta/core/terminal.h>
```
jviotti, line 1: As a small convention, can you keep this include first, a blank line, THEN the local include in quotes, THEN a blank line, then the rest?

src/core/terminal/terminal.cc
```diff
@@ -0,0 +1,157 @@
...
+    -> std::size_t {
+  switch (stream) {
+    case sourcemeta::core::TerminalStream::Stdin:
+      return 0;
+    case sourcemeta::core::TerminalStream::Stdout:
+      return 1;
+    case sourcemeta::core::TerminalStream::Stderr:
+      return 2;
+  }
+  return 1;
```
jviotti: Instead of a hardcoded default, you can use `default: ` with `std::unreachable()`. Thus making it impossible for the switch to go out of date

src/core/terminal/terminal.cc
```diff
@@ -0,0 +1,157 @@
...
+auto terminal_is_interactive(TerminalStream stream) noexcept -> bool {
+  return internal::is_interactive_stream(stream);
+}
+
+auto terminal_is_interactive(int file_descriptor) noexcept -> bool {
+  return internal::is_interactive_fd(file_descriptor);
+}
+
+auto terminal_set_color_policy(TerminalColorPolicy policy) noexcept -> void {
+  for (auto &stream_policy : g_stream_policies) {
```
jviotti: I think the `g_` prefix looks weird here, as we don't have that convention for globals anywhere else in the code

src/core/terminal/terminal.cc
```diff
@@ -0,0 +1,157 @@
...
+}
+
+auto terminal_color_policy(TerminalStream stream) noexcept
+    -> TerminalColorPolicy {
+  return g_stream_policies[stream_index(stream)].load(
+      std::memory_order_relaxed);
+}
+
+auto terminal_color_enabled(TerminalStream stream) noexcept -> bool {
+  if (terminal_color_policy(stream) == TerminalColorPolicy::Disabled) {
```
jviotti: This whole logic in this function could be a one liner with `&&` / `||` or whatever?

src/core/terminal/terminal.cc
```diff
@@ -0,0 +1,157 @@
...
+    return false;
+  }
+  return terminal_is_interactive(stream);
+}
+
+auto terminal_sgr_reset() noexcept -> std::string_view { return "\033[0m"; }
```
jviotti: Can we define these actual weird escape values in another header like `src/core/terminal/terminal_ansi_escapes.h` or something like that, as a collection of `constexpr std::string_view` constants with clear comments mapping them out to specs? Just to maintain them better there

src/core/terminal/terminal_posix.cc
```diff
@@ -0,0 +1,35 @@
+#include "terminal_internal.h"
+
+#include <unistd.h> // isatty, STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO
+
+namespace sourcemeta::core::internal {
+
+auto is_interactive_stream(TerminalStream stream) noexcept -> bool {
+  int file_descriptor{STDOUT_FILENO};
+  switch (stream) {
```
jviotti: Ahh if you do what I said on the public header, where you manually associate the enum class values to integer, I think you can autoamtically convert to them using `std::underlying_type` or something like that (don't recall the name), thus avoiding the switch

src/core/terminal/terminal_posix.cc
```diff
@@ -0,0 +1,35 @@
...
+}
+
+auto is_interactive_fd(int file_descriptor) noexcept -> bool {
+  if (file_descriptor < 0) {
+    return false;
+  }
+  return ::isatty(file_descriptor) == 1;
+}
+
+auto enable_virtual_terminal_stream(TerminalStream /*stream*/) noexcept
```
jviotti: I don't think you need the `/*stream*/` comment? the fact the type says `Stream` already makes it clear?

test/terminal/terminal_test.cc
```diff
@@ -0,0 +1,272 @@
...
+
+TEST(stream_detection_runs_safely) {
+  // Exercises stream queries without throws or crashes
+  const bool stdin_is_term{sourcemeta::core::terminal_is_interactive(
+      sourcemeta::core::TerminalStream::Stdin)};
+  const bool stdout_is_term{sourcemeta::core::terminal_is_interactive(
+      sourcemeta::core::TerminalStream::Stdout)};
+  const bool stderr_is_term{sourcemeta::core::terminal_is_interactive(
+      sourcemeta::core::TerminalStream::Stderr)};
+  static_cast<void>(stdin_is_term);
```
jviotti: This looks off. Instead of casting to `void`, you can use `[[maybe_unused]]`. That is the C++ idiomatic way!

test/terminal/terminal_test.cc
```diff
@@ -0,0 +1,272 @@
...
+                                   sourcemeta::core::TerminalStyle::Red, false);
+  EXPECT_EQ(stream_output.str(), "Plain");
+
+  std::ostringstream none_stream_output;
+  sourcemeta::core::terminal_paint(none_stream_output, "NoneStyle",
+                                   sourcemeta::core::TerminalStyle::None, true);
+  EXPECT_EQ(none_stream_output.str(), "NoneStyle");
+}
+
+TEST(style_bitwise_operators) {
```
jviotti: Can you have separate tests for each operator? In general, more decoupled tests are better than single ones that do a lot

test/terminal/terminal_test.cc
jviotti, line 1: Are we exercising all functions here? Might be worth having at least one case exercising the global style?

src/core/terminal/include/sourcemeta/core/terminal.h
jviotti: i.e. you could even unit test that you can seamlessly convert `STDIN_FILENO` from `<stdio.h>` into it :)

src/core/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -0,0 +1,498 @@
...
+    ++color_count;
+  }
+  if ((style & TerminalStyle::Blue) != TerminalStyle::None) {
+    ++color_count;
+  }
+  if ((style & TerminalStyle::Cyan) != TerminalStyle::None) {
+    ++color_count;
+  }
+
+  return color_count <= 1;
```
jviotti: A bit minor, but by how we are accumulating on the color count, we could probably break early already whenever there are two bumps?

src/core/terminal/terminal.cc
```diff
@@ -0,0 +1,171 @@
...
+// Indices match TerminalStream: Stdin = 0, Stdout = 1, Stderr = 2
+std::array<std::atomic<sourcemeta::core::TerminalColorPolicy>, 3>
+    stream_policies{{sourcemeta::core::TerminalColorPolicy::WhenInteractive,
+                     sourcemeta::core::TerminalColorPolicy::WhenInteractive,
+                     sourcemeta::core::TerminalColorPolicy::WhenInteractive}};
+
+constexpr auto stream_index(sourcemeta::core::TerminalStream stream) noexcept
+    -> std::size_t {
+  switch (stream) {
+    case sourcemeta::core::TerminalStream::Stdin:
```
jviotti: Can't we use https://en.cppreference.com/cpp/utility/to_underlying?

src/core/terminal/terminal_ansi_escapes.h
```diff
@@ -0,0 +1,56 @@
+#ifndef SOURCEMETA_CORE_TERMINAL_ANSI_ESCAPES_H_
+#define SOURCEMETA_CORE_TERMINAL_ANSI_ESCAPES_H_
+
+#include <string_view> // std::string_view
+
+namespace sourcemeta::core::internal {
+
+/// @name ECMA-48 Select Graphic Rendition (SGR) Control Sequences
```
jviotti: No need to use Doxygen syntax here. These are internal headers that are never run through Doxygen. Just use normal comment blocks here!

src/core/terminal/include/sourcemeta/core/terminal.h
HarshPopat23: Good point! Updated `terminal_style_is_valid` to iterate through the foreground colors using a `constexpr std::array` and return `false` immediately as soon as a second color flag is encountered (`if (++color_count > 1) return false;`), avoiding unnecessary checks.

src/core/terminal/terminal.cc
HarshPopat23: Since we are on C++23, `std::to_underlying` from `<utility>` is the perfect fit here. Replaced the switch statement with:

```cpp
constexpr auto stream_index(sourcemeta::core::TerminalStream stream) noexcept
    -> std::size_t {
  return static_cast<std::size_t>(std::to_underlying(stream));
}

src/core/terminal/terminal_ansi_escapes.h
HarshPopat23: Done. Stripped all Doxygen tags (`///`, `@name`, `@{`, `@}`, `@see`) from this internal header and replaced them with standard `//` comment blocks.

jviotti (approved):

PR #2808, opened 2026-09-11
feat(terminal): Add an unconditional terminal colour policy

HarshPopat23:
### Summary

Captured output is noninteractive, but callers sometimes explicitly request ANSI styling. `TerminalColorPolicy::Always` supports that use case while preserving automatic detection (`WhenInteractive`) and disabled styling (`Disabled`). It also enables downstream CLI tests to assert ANSI output through ordinary capture (e.g. `.clitest` runners).

### Changes

- Added `TerminalColorPolicy::Always` to `sourcemeta/core/terminal.h`.
- Updated `terminal_color_enabled` in `terminal.cc` to evaluate all three policies:
  - `WhenInteractive`: preserves destination-based terminal detection.
  - `Always`: unconditionally enables styling for all destinations (including pipes and redirected files).
  - `Disabled`: unconditionally suppresses styling.
- Kept `terminal_is_interactive` truthful: `Always` does not alter terminal detection.
- Updated public documentation and examples in `terminal.h`.
- Extended the terminal test suite in `terminal_test.cc`:
  - Added RAII `ScopedColorPolicy` to isolate and restore policy state across test cases.
  - Added tests for `Always`, `Disabled`, and `WhenInteractive` policy behavior and transitions.
  - Added tests for destination-aware `terminal_paint` overloads (string and ostream) under `Always` and `Disabled`.
  - Added noninteractive child process tests verifying exact ANSI output under `Always` and plain text under `WhenInteractive`/`Disabled` when output is captured through pipes.

<a href="https://cubic.dev/pr/sourcemeta/core/pull/2808?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/core/terminal/include/sourcemeta/core/terminal.h
```diff
@@ -351,6 +359,15 @@ auto terminal_color_policy(
 ///
 /// Determine whether styling is enabled for the specified stream.
 ///
+/// Under `TerminalColorPolicy::WhenInteractive`, styling is enabled if the
```
jviotti: I feel these doc additions are just re-explaining what you documented in the enum class? Can we remove them?

test/terminal/CMakeLists.txt
```diff
@@ -3,3 +3,10 @@ sourcemeta_test(NAMESPACE sourcemeta PROJECT core NAME terminal
 
 target_link_libraries(sourcemeta_core_terminal_unit
   PRIVATE sourcemeta::core::terminal)
+
+if(SOURCEMETA_CORE_LANG_PROCESS)
```
jviotti: Why make it conditional though, with the corresponding conditional preprocessor checks in the test file? We can just make it a required dependency of this to avoid the extra complexity?

test/terminal/terminal_test.cc
```diff
@@ -311,3 +377,339 @@ TEST(composition_with_formatting) {
...
+  // Under Disabled, child connected to pipe produces plain output
+  const auto disabled_result{sourcemeta::core::spawn_and_capture(
+      helper, {"--child-noninteractive-disabled"})};
+  EXPECT_TRUE(disabled_result.exit_code.has_value());
+  EXPECT_EQ(disabled_result.exit_code.value(), 0);
+  EXPECT_EQ(disabled_result.standard_output, "streamed\npainted");
+#endif
+}
+
+auto main(int argc, char **argv) -> int {
```
jviotti: We shouldn't have a single test file doing both `TEST()` invocations AND a `main` entry point. As the `sourcemeta/core/test.h` runner IS providing its own entry point to execute the other ones.

I think a better design here is to split this test CLI into a NEW .cc file in `test/terminal`. Then this file can spawn it on its test logic, or you can hook it directly over CMake's `add_test`?

It also raises an interesting point: here you are trying to test an actual CLI for this. Which means: why can't we use `.clitest` in this module itself for this? Sounds like the perfect dog-fooding?

test/terminal/CMakeLists.txt
```diff
@@ -2,4 +2,16 @@ sourcemeta_test(NAMESPACE sourcemeta PROJECT core NAME terminal
   SOURCES terminal_test.cc)
 
 target_link_libraries(sourcemeta_core_terminal_unit
+  PRIVATE sourcemeta::core::terminal
+  PRIVATE sourcemeta::core::process)
+
+sourcemeta_executable(NAMESPACE sourcemeta PROJECT core NAME terminal
+  VARIANT unit_child_main SOURCES terminal_child_main.cc)
```
jviotti:
```suggestion
  VARIANT test_helper SOURCES terminal_child_main.cc)
```

Nicer name?

test/terminal/CMakeLists.txt
```diff
@@ -2,4 +2,16 @@ sourcemeta_test(NAMESPACE sourcemeta PROJECT core NAME terminal
...
+target_link_libraries(sourcemeta_core_terminal_unit_child_main
   PRIVATE sourcemeta::core::terminal)
+
+add_dependencies(sourcemeta_core_terminal_unit
+  sourcemeta_core_terminal_unit_child_main)
+target_compile_definitions(sourcemeta_core_terminal_unit
+  PRIVATE HELPER_BINARY="$<TARGET_FILE:sourcemeta_core_terminal_unit_child_main>")
+
```
jviotti, line 19: Unexpected empty line?

test/terminal/terminal_child_main.cc
```diff
@@ -0,0 +1,54 @@
...
+                       sourcemeta::core::TerminalStyle::Green)
+            << std::flush;
+  return 0;
+}
+
+} // namespace
+
+auto main(int argc, char **argv) -> int {
+#if defined(_WIN32)
+  _setmode(_fileno(stdout), _O_BINARY);
```
jviotti, line 40: Why is this needed btw? Might be worth a comment explaining it?

test/terminal/terminal_child_main.cc
```diff
@@ -0,0 +1,54 @@
...
+#endif
+
+  if (argc > 1) {
+    const std::string_view mode{argv[1]};
+    if (mode == "--child-noninteractive-when-interactive") {
```
jviotti: At least for encouraging re-use, it might be worth using `src/lang/options` to parse CLI options. That's the CLI option parser we have in this project

test/terminal/terminal_test.cc
```diff
@@ -12,6 +14,61 @@
 #include <unistd.h> // pipe, close
 #endif
 
+namespace {
+
+class ScopedColorPolicy {
```
jviotti: Interesting. I think the fact we need a complex RAII guard for this makes me think whether we should have a proper `_reset()` function in `src/core/terminal`? Then the module, inside of itself, can always figure out how to fallback to the original policy?

test/terminal/terminal_test.cc
```diff
@@ -311,3 +371,275 @@ TEST(composition_with_formatting) {
...
+            "Info");
+
+  std::ostringstream stream_output;
+  sourcemeta::core::terminal_paint(stream_output,
+                                   sourcemeta::core::TerminalStream::Stdout,
+                                   "Streamed", bold_blue);
+  EXPECT_EQ(stream_output.str(), "Streamed");
+}
+
+TEST(noninteractive_child_execution) {
```
jviotti: I still think replacing this test case with a couple of real `.clitest` files in test/terminal and registering them like that in `test/terminal/CMakeLists.txt` would be nicer

HarshPopat23:
Thank you @jviotti  for the review! I have addressed all feedback.
PTAL.

jviotti (approved):
Wonderful! Well done

